PR #54, opened 2015-01-11
Definition file for WilliamChart library included

diogobernardino:


mikepenz:
@diogobernardino thank you

PR #63, opened 2015-02-16
New Libraries and fix for text from Library A appearing on Library B's card

jkane001:
I'm including 2 changes: first, 5 new libraries. None are open source, so there is no direct way to get the authors to include these strings (especially since these are owned by Google, Twitter, and Facebook). Best guesses are made for a couple of these, as well, though all information should be accurate.

Second change is one to solve a problem I noted: I had left version numbers out of these 5 library's data, but other libraries had version number data. With RecyclerView and the ViewHolder pattern, as soon as an item is scrolled off-screen, it gets removed from memory. When you scroll back, those cards are re-built using the ViewHolder in memory, and thus you get version numbers from other libraries showing on the cards of libraries that don't have version data. Removing static from the ViewHolder should resolve that.

### DIFF
--- a/library/src/main/java/com/mikepenz/aboutlibraries/ui/adapter/LibsRecyclerViewAdapter.java
+++ b/library/src/main/java/com/mikepenz/aboutlibraries/ui/adapter/LibsRecyclerViewAdapter.java
@@ -247,7 +247,7 @@ public class LibsRecyclerViewAdapter extends RecyclerView.Adapter<RecyclerView.V
         this.header = false;
     }
 
-    public class HeaderViewHolder extends RecyclerView.ViewHolder {
+    public static class HeaderViewHolder extends RecyclerView.ViewHolder {
         ImageView aboutIcon;
         TextView aboutAppName;
         TextView aboutVersion;
@@ -266,7 +266,7 @@ public class LibsRecyclerViewAdapter extends RecyclerView.Adapter<RecyclerView.V
         }
     }
 
-    public class ViewHolder extends RecyclerView.ViewHolder {
+    public static class ViewHolder extends RecyclerView.ViewHolder {
         View card;
 
         TextView libraryName;
--- a/library/src/main/java/com/mikepenz/aboutlibraries/ui/adapter/LibsRecyclerViewAdapter.java
+++ b/library/src/main/java/com/mikepenz/aboutlibraries/ui/adapter/LibsRecyclerViewAdapter.java
@@ -247,7 +247,7 @@ public class LibsRecyclerViewAdapter extends RecyclerView.Adapter<RecyclerView.V
         this.header = false;
     }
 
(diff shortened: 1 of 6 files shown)


mikepenz:
Yeah i've looked over the 5 new libraries, they seem accurate. I'm still not sure if it's required or necessary to include the google support libraries, but it's nice to have a definition file for those. In fact they won't be auto-detected, and you have to pass them in the builder.

```java
.withLibraries("AppCompat", "GooglePlayServices")
```


Talking about the second change. I see your point, but the fix you've provided is not correct. The ViewHolder itself should be static that's fine. The issue is somewhere else. 
The problem is in the ```onBindViewHolder``` method

To be correct it's in this part of the method.

```java
//Set License or Version Text
            if (TextUtils.isEmpty(library.getLibraryVersion()) && library.getLicense() != null && TextUtils.isEmpty(library.getLicense().getLicenseName()) || (!showVersion && !showLicense)) {
                holder.libraryBottomDivider.setVisibility(View.GONE);
                holder.libraryBottomContainer.setVisibility(View.GONE);
            } else {
                holder.libraryBottomDivider.setVisibility(View.VISIBLE);
                holder.libraryBottomContainer.setVisibility(View.VISIBLE);

                if (!TextUtils.isEmpty(library.getLibraryVersion()) && showVersion) {
                    holder.libraryVersion.setText(library.getLibraryVersion());
                }
                if (library.getLicense() != null && !TextUtils.isEmpty(library.getLicense().getLicenseName()) && showLicense) {
                    holder.libraryLicense.setText(library.getLicense().getLicenseName());
                }
            }
```

as you can see, i check if the version is not empty, if it is i won't set the new text. I have to add an else block like this:

```java
else {
   holder.libraryVersion.setText("");
}
```

jkane001:
Ah, good call on the bug fix.

As far as the auto-detection, it worked for me without having to explicitly
call out those libraries, so I'm not sure what you mean.

I also an not sure it's necessary to list those libraries on the about
page, but seems better to err on the side of caution, what with Google's
propensity toward pulling apps from the store with no warning.

BTW, I spent some time last night looking into auto-detecting version
numbers, and it seems it's not really possible, so I guess that feature
request is dead in the vine. :)

Thanks,
J

mikepenz:
@jkane001 can you modify the pull request? so i can include half of it ;)

yeah you are right, totally missed that you've also included the classPath. 
sorry for the misunderstanding.


yeah you are right!. 


the only possibility to include the version would be to write a gradle plugin which would include the version within the definition file or something like this. i've also spent a lot of time for this one already. and yeah. at the moment i'm afraid i have not the time to do something like this. 

Thanks a lot!

jkane001:
Creating new PR without the bug fix

mikepenz:
@jkane001 ok thanks ;)

i'm still not really sure how to handle those libs. somehow the aosp libs are "normal" in each app :D

PR #74, opened 2015-03-25
Added new libraires

cyroxis:


library/src/main/res/values/library_sugarorm_strings.xml
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="define_sugarorm"></string>
```
mikepenz: please add `_int` to the define name. or it will always show up. even if the library is not used.

mikepenz:
@cyroxis thanks. sorry for the late merge. was a bit busy with the MaterialDrawer library

PR #88, opened 2015-05-10
2 new Libraries

RubenGees:
Added these Libraries I use in my Project:
https://github.com/lecho/hellocharts-android
https://github.com/ApmeM/android-flowlayout

### DIFF
--- a/library/src/main/res/values/library_androidflowlayout_strings.xml
+++ b/library/src/main/res/values/library_androidflowlayout_strings.xml
@@ -1,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<resources>
+	<string name="define_int_androidflowlayout"></string>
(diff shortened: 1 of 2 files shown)


mikepenz:
@RubenGees thanks

RubenGees:
Those Libraries don't show up. Any idea why?

mikepenz:
@RubenGees yeah i just saw that you set the classPath value wrong, so it won't automatically show up. 

The classPath value should be a specific class which is inside the library. Like org.apmem.tools.layouts.FlowLayout for the android-flowlayout library

RubenGees:
Can it be any random class from the lib?

mikepenz:
@RubenGees it can be. but should one which is one of the Main classes so the chance is very low that it will be removed

PR #90, opened 2015-05-21
Dagger and Dagger 2

wKovacs64:
* Separate Dagger detection from Dagger 2
* Add internal support for Dagger 2


mikepenz:
thank you ;)

PR #91, opened 2015-05-28
MPAndroidChart

nanopc:
Added MPAndroidChart 
https://github.com/PhilJay/MPAndroidChart

### DIFF
--- a/library/src/main/res/values/library_mpandroidchart_strings
+++ b/library/src/main/res/values/library_mpandroidchart_strings
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 
 <resources>
-	<string name="define_mpandroidchart"></string>
+	<string name="define_int_mpandroidchart"></string>
 	<!-- Author section -->
 	<string name="library_mpandroidchart_author">PhilJay</string>
 	<string name="library_mpandroidchart_authorWebsite">https://github.com/PhilJay/MPAndroidChart</string>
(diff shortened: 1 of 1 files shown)


library/src/main/res/values/library_mpandroidchart_strings
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<resources>
+	<string name="define_mpandroidchart"></string>
```
mikepenz: it should say define_int_mpandroidchart

(The int defines that this library is inside of the AboutLibraries library and should not be displayed automatically if the xml is found, only if the user chooses to show it)

library/src/main/res/values/library_mpandroidchart_strings
```diff
@@ -0,0 +1,21 @@
...
+	<!-- Library section -->
+	<string name="library_mpandroidchart_libraryName">MPAndroidChart</string>
+	<string name="library_mpandroidchart_libraryDescription">MPAndroidChart is a powerful and easy to use chart library for Android, supporting line-, bar-, scatter-, candlestick-, bubble-, pie- and radarcharts (spider web), as well as scaling, dragging (panning), selecting and animations.</string>
+	<string name="library_mpandroidchart_libraryWebsite">https://github.com/PhilJay/MPAndroidChart</string>
+	<string name="library_mpandroidchart_libraryVersion">v2.0.8</string>
+	<!-- OpenSource section -->
+	<string name="library_mpandroidchart_isOpenSource">true</string>
+	<string name="library_mpandroidchart_repositoryLink">https://github.com/PhilJay/MPAndroidChart</string>
+	<!-- ClassPath for autoDetect section -->
+	<string name="library_mpandroidchart_classPath"></string>
```
mikepenz: You should also define a classPath. This is a class (the whole packagename inclusive class name) which is one of the main classes inside the library. So the AboutLibraries library can autodetect this one.

mikepenz:
hey. thank you so much.

mikepenz:
@nanopc great ;)
thank you so much

nanopc:
Thanks to you! ;)

PR #92, opened 2015-05-28
fixes using repositoryLink if libraryWebsite is not set

Fentin:
without the fix, this would not make sense:

library.getLibraryWebsite() != null ? library.getLibraryWebsite() : library.getRepositoryLink()

;-)


mikepenz:
@Fentin you are totally right. Thank you so much

PR #108, opened 2015-06-18
Allow specification of a comparator to control sorting of the libraries.

jgindin:


mikepenz:
@jgindin thanks

jgindin:
When do you think you'll release a new build?

mikepenz:
after i tried everything
perhaps this weekend

mikepenz:
@jgindin ok published :D

PR #115, opened 2015-06-25
Make sure the builder's comparator get used in all cases. 

jgindin:
Changes made ensure to preserve the existing API and behaviors.

Added sample activity to show usage of custom sorting.


mikepenz:
@jgindin thanks. looks good.

mikepenz:
@jgindin it would be great if you could create the pr on the develop branch

jgindin:
Would you like me to do that for this review? Or just for future? Either way works for me.

(And, I'll take this opportunity to thank you for some great libraries!)

mikepenz:
@jgindin for this please. i use git flow so develop is the branch for development ;)
i changed the default branch already so it should not happen anymore then :D

No problem ;). Thank you

PR #117, opened 2015-06-29
Make sure the builder's comparator get used in all cases.

jgindin:
Changes made ensure to preserve the existing API and behaviors.
Added sample activity to show usage of custom sorting.

NOTE: Before adding my commit for this functionality (e9f2786) I had to first merge from master into develop. I admit to being relatively inexperienced with git, so please let me know if there's a better / more preferred way of doing this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ You can find anything you search for in the wiki. (If not open an issue)
 The AboutLibraries Library is pushed to [Maven Central](http://search.maven.org/#search|ga|1|g%3A%22com.mikepenz%22), so you just need to add the following dependency to your `build.gradle`. It seems it is also required to add the support dependencies to the application. If it works without, you should be fine too :).
 
 ```javascript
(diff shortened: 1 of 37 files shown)


mikepenz:
@jgindin it would be easier if you revert the changes on the develop branch. and just cherry pick the single change. merging this pull request would be the same as me merging the master into develop which is not a good thing ;)

jgindin:
I tried that, but since there are changes on master which are not in the develop branch (such as the last commit I made, adding the libraryComparator field to LibsBuilder) it didn't work.

I think I'm just missing something about this workflow... can you point me at some reading? I'll go figure it out and then submit a new pull request.

Sorry for the gyrations here...

mikepenz:
@jgindin to simplify the whole thing i will just cherry pick the change from your branch. for the next time ;). pull the develop branch first and then merge or commit the changes ;)

PR #127, opened 2015-07-30
10 new libraries

JordanGout:
Hope everything is correct.

### DIFF
--- a/library/src/main/res/values/library_androidfloatingactionbutton_strings.xml
+++ b/library/src/main/res/values/library_androidfloatingactionbutton_strings.xml
@@ -1,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<resources>
+	<string name="define_int_android_floating_action_button"></string>
+	<!-- Author section -->
+	<string name="library_android_floating_action_button_author">Jerzy Chalupski</string>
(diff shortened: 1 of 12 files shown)


mikepenz:
@JordanGout it seems fine. Just one thing. The classpaths you have defined should point to an class which can be instantiated. Else it's not really easy for the lib to make sure a specific lib exists. (I use this to see if the class which is written can be found and if yes i'll automatically add the lib) 

Can you please change it to point to a real class? I will do a quick check on something i just had in mind. i'll let you know if it worked ;)

mikepenz:
@JordanGout ok the idea i had seems to be not possible. :/

JordanGout:
I've changed the classpaths to a real class.

I also changed the classpaths of crashlytics and facebook because there are out of date.

mikepenz:
@JordanGout looks great :) thank you

vanniktech:
Hmm I can't find the `com.facebook.FacebookSdk` class. I'm using `'com.facebook.android:facebook-android-sdk:3.22.0'` Is `FacebookSdk` only present in 4 and above?

mikepenz:
@vanniktech if the auto detection does not work you can also just enable it via the builder. `withLibraries("facebook")` perhaps the classpath is wrong and should be changed? you can open a pull request if you like to ;)

vanniktech:
I found one issue, you'll get a PR in a few moments.

PR #134, opened 2015-08-27
Fixes issue that some libraries were not detected

vanniktech:
I don't know why but you can't check for 0 here it has to be > 0. Also I'm not quite sure whether this one will introduce bugs that I'm not aware of. I just quickly debugged and found out that this is the sweet spot where things go wrong.

To verify you can add the facebook library in `sample/build.gradle` with `compile 'com.facebook.android:facebook-android-sdk:3.22.0'`

Prior to this commit it was not detected. After this commit it'll be detected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ You can find anything you search for in the wiki. (If not open an issue)
 The AboutLibraries Library is pushed to [Maven Central](http://search.maven.org/#search|ga|1|g%3A%22com.mikepenz%22), so you just need to add the following dependency to your `build.gradle`. It seems it is also required to add the support dependencies to the application. If it works without, you should be fine too :).
 
 ```javascript
-compile('com.mikepenz:aboutlibraries:5.2.0@aar') {
+compile('com.mikepenz:aboutlibraries:5.2.1@aar') {
 	transitive = true
 }
 ```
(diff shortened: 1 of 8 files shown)


mikepenz:
@vanniktech thanks for the pull request. but i can't merge as this will break the caching of the auto detected libs. 

Can you retry if it works for you without this change? to make sure it does not listen for the cache just count up the versionCode of your app

vanniktech:
I tried it out with the sample app, so you're saying that I should just increase the versionCode and remove the 'fix'?

mikepenz:
the pull request is no fix. it just disables the cache which remembers the auto detected libs. this only happens once per versionCode (to make it faster) 

so if you change definition files (or update the lib) you will have to remove app data or count the version up that new libs are detected

vanniktech:
I changed the versionCode in the sample library to 521 with the facebook change and it still did not recognize the library.

mikepenz:
@vanniktech i looked over the code and it should do exactly what i wrote above. it works for you just with the change (the pullrequest) and no other changes?

vanniktech:
Yes exactly. Obviously you have to add the Facebook SDK to the sample project though.

`compile 'com.facebook.android:facebook-android-sdk:3.22.0'`

mikepenz:
@vanniktech just checked. somehow the FacebookSdk class is not inside the lib itself. It will work if i change it to: `com.facebook.FacebookSdkVersion`

vanniktech:
Oh okay nice you changed it. It does work indeed now.

PR #137, opened 2015-08-29
Add PixCell8 to apps using AboutLibraries

wKovacs64:


mikepenz:
and another thanks :D

PR #140, opened 2015-09-02
Adds ddplist library

vanniktech:
I've got one question though when I import the library itself in the sample project with `compile 'com.googlecode.plist:dd-plist:1.3'` in `sample/build.gradle` it does not automatically show up.

I have to go in [FragmentActivity line 170](https://github.com/mikepenz/AboutLibraries/blob/develop/sample/src/main/java/com/mikepenz/aboutlibraries/sample/FragmentActivity.java#L170) and add it manually with `.withLibraries("crouton", "activeandroid", "actionbarsherlock", "showcaseview", "ddplist")`. 

Did I do something wrong or is it just the way the library works?

Also is the library able to run properly and show every library even though when in release proguard is turned on and packages may get renamed?

### DIFF
--- a/library/src/main/res/values/library_ddplist_strings.xml
+++ b/library/src/main/res/values/library_ddplist_strings.xml
@@ -1,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="define_int_ddplist">year;owner</string>
+    <string name="library_ddplist_author">Daniel Dreibrodt</string>
+    <string name="library_ddplist_libraryName">dd-plist</string>
+    <string name="library_ddplist_libraryDescription">
+        <![CDATA[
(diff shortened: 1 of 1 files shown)


mikepenz:
@vanniktech there is an internal cache which keeps the libraries which were detected via the auto detection per `versionCode` so you will have to count up. If you do this it will show up. 

If you enable proguard and packagenames of libraries are altered the auto detection won't work for libs which are included in this repository. It will only work for libraries which include the definition file itself (like MaterialDrawer, etc...) This is because the only thing i have found working was to check for a specific Class of a lib to check if it is there. 

In this case i recommend you to just provide those libs manually. Which shouldn't be to much extra work.

vanniktech:
Alright thank you very much. If I find the time, I'll add some other new libraries to this project as well.

PR #142, opened 2015-09-03
Adds intellijannotations library

vanniktech:


mikepenz:
Are you sure this makes sense and will show up? This is just a IDEA plugin which will help you during your development process, but it won't do much for the real app you release? As far as i understand it those are just annotations?

vanniktech:
Yup those are just annotations but as far as I know you still have to include their license if you use them during development. Also I think they do show up in the final dex file and don't get compiled away.

PR #157, opened 2015-09-22
BetterPickers lib has moved and is now maintained by Code-Troopers

fchauveau:
https://github.com/code-troopers/android-betterpickers

PR #158, opened 2015-09-25
Update BetterPickers aboutLib infos / second try

fchauveau:
Same PR than #157 but on the correct branch this time.

Thx


mikepenz:
thank you so much ;)

PR #172, opened 2015-11-05
Update internal Butter Knife data

wKovacs64:
Just some housekeeping.


mikepenz:
@wKovacs64 thank you.

PR #181, opened 2015-12-22
Added magister.java

iLexiconn:


mikepenz:
@iLexiconn you are the author of that lib?

iLexiconn:
Yes I am. :)

mikepenz:
@iLexiconn you can also add this definition file to your library, so it will automatically show up for people using the `AboutLibraries`. I do the same for my libs:
https://github.com/mikepenz/MaterialDrawer/blob/develop/library/src/main/res/values/library_materialdrawer_strings.xml

just change ** define_int_magisterjava** to ** define_magisterjava**

iLexiconn:
The library is not made in an Android development environment, so I can't do it that way.

mikepenz:
ah ok i see. then thanks for your contribution. will be in the next version

iLexiconn:
:)

PR #211, opened 2016-04-11
Add LoganSquare config

Jawnnypoo:
Its a great Gson alternative!
https://github.com/bluelinelabs/LoganSquare

### DIFF
--- a/library/src/main/res/values/library_logansquare_strings.xml
+++ b/library/src/main/res/values/library_logansquare_strings.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="define_int_LoganSquare">year;owner</string>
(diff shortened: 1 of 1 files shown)


mikepenz:
@Jawnnypoo oh thank you

library/src/main/res/values/library_logansquare_strings.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="define_int_LoganSquare">year;owner</string>
+    <string name="library_AboutLibraries_author">BlueLine Labs, Inc.</string>
```
mikepenz: you have a copy paste error here. should be library_LoganSquare too

Jawnnypoo:
Good catch on Copy paste. Should be good now!

mikepenz:
Looks fine now. Thanks

PR #221, opened 2016-06-22
Multiple Library Updates

WoogieNoogie:
Added Mixpanel and RxLifecycle rules.
Made modifications to the RxJava, RxAndroid, and SugarORM rules.


library/src/main/res/values/library_mixpanel_strings.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<resources>
+    <string name="define_mixpanelandroid"></string>
```
mikepenz: should be define_int_* as like here:
https://github.com/WoogieNoogie/AboutLibraries/blob/19167939b4ffccb09ada141d1dd46904e96d9d53/library/src/main/res/values/library_sugarorm_strings.xml#L3

library/src/main/res/values/library_rxlifecycle_strings.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<resources>
+    <string name="define_rxlifecycle"></string>
```
mikepenz: should be define_int_* as like here:
https://github.com/WoogieNoogie/AboutLibraries/blob/19167939b4ffccb09ada141d1dd46904e96d9d53/library/src/main/res/values/library_sugarorm_strings.xml#L3

mikepenz:
@WoogieNoogie which odd issues are you talking about?

WoogieNoogie:
Mainly, I've included SugarORM and RxJava in my project, but neither are appearing in the about page. Hopefully the changes I made have fixed this, but at the very least, these two new libraries can be included. I've made the fixes from your comments above.

mikepenz:
SugarORM is not showing up because it has a wrong classPath here:
https://github.com/mikepenz/AboutLibraries/pull/221/files#diff-2e6006abc2fa8c46012f43afc41c259aR16

Didn't see it before. 

The problem is that I have to try to init the class to find out if a specific library is included. So therefor I introduced the `classPath`. It may has a bit a wrong name. 

As an alternative the files can be included in the libs themself just having the `define_` prefix. This will tell the aboutlibs that the library is used and that the resource comes from the library and not from the AboutLibraries. 

So in case of SugarOrm it should be like:
https://github.com/mikepenz/AboutLibraries/pull/221/files#diff-2e6006abc2fa8c46012f43afc41c259aR16
`com.orm.SugarApp`

WoogieNoogie:
Okay, maybe I didn't 100% understand how this library pulled in the other projects. It can't just use the package path, it needs to be a full path to a class that's used in the project - is that right?

mikepenz:
Yes correctly. There is no way I have found which allows me to check for a specific package to exist

WoogieNoogie:
Okay great, sorry for the confusion. Let me fix things up a bit and I'll comment when things are looking better.

mikepenz:
@WoogieNoogie sieht gut aus.

mikepenz:
@WoogieNoogie Oh sorry that was the wrong language. Looks fine ;)

PR #223, opened 2016-07-05
Foundation Icons License

Rainer-Lang:


mikepenz:
@Rainer-Lang please add the definition file to the Android Iconics. To the specific font:
https://github.com/mikepenz/Android-Iconics

(don't forget to remove the "_int" from the define part)

PR #228, opened 2016-07-18
Add AppRate strings

MSF-Jarvis:
This should be mergeable now :+1:

### DIFF
--- a/library/src/main/res/values/library_apprate_strings
+++ b/library/src/main/res/values/library_apprate_strings
@@ -1,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<resources>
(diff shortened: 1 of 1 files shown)


mikepenz:
@MSF-Jarvis just saw that this is your library. You can include this file into your library, with a single change. Change `define_int_apprate` to `define_apprate`. So it will automatically show up as soon as someone uses your lib

MSF-Jarvis:
:slaphead:

I'll get it done. PR closed

mikepenz:
@MSF-Jarvis it makes more sense for you if it's in your library, as you can keep it up2date with your library getting updated. and it will show up automatically inside the aboutlibs

PR #243, opened 2016-08-12
Updated Jsoup information and added ProcessPhoenix

Namnodorel:
Updated the version and license information of Jsoup and added the ProcessPhoenix library


mikepenz:
@Namnodorel thank you

PR #245, opened 2016-08-16
Use application Context with LibraryTask.

ronaldsmartin:
We don't specifically need the activity context in the operation.


mikepenz:
thanks :)

PR #253, opened 2016-09-01
New libs

rubengees:
This PR covers

- [Android TextView Builder](https://github.com/klinker24/Android-TextView-LinkBuilder)
- [Anko](https://github.com/Kotlin/anko)
- [Kotterknife](https://github.com/JakeWharton/kotterknife)
- [Moshi](https://github.com/square/moshi)

which I use in one of my projects.

### DIFF
--- a/library/src/main/res/values/library_android_textview_linkbuilder_strings.xml
+++ b/library/src/main/res/values/library_android_textview_linkbuilder_strings.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 
 <resources>
-    <string name="define_android_textview_linkbuilder"></string>
+    <string name="define_int_android_textview_linkbuilder"></string>
     <!-- Author section -->
     <string name="library_android_textview_linkbuilder_author">Luke Klinker</string>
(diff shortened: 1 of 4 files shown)


library/src/main/res/values/library_android_textview_linkbuilder_strings.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<resources>
+    <string name="define_android_textview_linkbuilder"></string>
```
mikepenz: please add `_int` to the define definition. 
without `_int` those libs will always be displayed. 


It should look like this: https://github.com/mikepenz/AboutLibraries/blob/develop/library/src/main/res/values/library_activeandroid_strings.xml#L4

Please also do this for the other 2 libs

rubengees: Ah! Forget about those, fixed ;)

mikepenz:
thank you

rubengees:
I added another lib which is use, but forget to initially include in this PR.

mikepenz:
Oh ok thanks.

PR #256, opened 2016-09-13
Update build.gradle

pedronveloso:
Google broke the dependency on support libraries `24.2.0`, so updating to the latest will fix it. No other changes needed.


mikepenz:
thank you

PR #261, opened 2016-10-16
Fixes failing build

leofarage:
Updates entry for `build-tools` 24.0.1 as travis is crashing because of missing build-tools


mikepenz:
thank you

Issue #266, opened 2016-11-08
Different text color in library cards and app description


fettlaus:
I'm using `about_libraries_text_openSource` to change the color of my app description. Unfortunately `HeaderItem` and `LibraryItem` both use this attribute to style their text color. 

I need to set different colors on these texts because my Activity uses a dark background and I'd like to keep the library cards in the RecyclerView rather bright.

Is there some other way I just can't think of right now?

mikepenz:
The only solution would be to copy and paste the layouts for those items to your project:
https://github.com/mikepenz/AboutLibraries/blob/develop/library/src/main/res/layout/listheader_opensource.xml
https://github.com/mikepenz/AboutLibraries/blob/develop/library/src/main/res/layout/listitem_opensource.xml

and modify the color's for these texts. 

The build will automatically overwrite the layouts from the library when the apk is created

fettlaus:
That is not possible, because the text color gets overwritten every time by 
`aboutAppDescription = (TextView) headerView.findViewById(R.id.aboutDescription);
            aboutAppDescription.setTextColor(UIUtils.getThemeColorFromAttrOrRes(headerView.getContext(), R.attr.about_libraries_text_openSource, R.color.about_libraries_text_openSource));`
in HeaderItem.

PR #267, opened 2016-11-09
made the description styleable

fettlaus:
Added the possiblity to change the text color of the description. This fixes my problem mentioned in #266 while maintaining backwards compatibility.

In the long term, `about_libraries_dividerDark_openSource` and `about_libraries_dividerLight_openSource` may be renamed as well, to reflect their usage better. One is used exclusively in HeaderItem the other one in LibraryItem. Maybe they should be renamed to `about_libraries_divider_openSource` and `about_libraries_divider_description`.

### DIFF
--- a/library/src/main/res/values/attrs.xml
+++ b/library/src/main/res/values/attrs.xml
@@ -4,6 +4,9 @@
     <declare-styleable name="AboutLibraries">
         <attr name="about_libraries_window_background" format="color|reference"/>
         <attr name="about_libraries_card" format="color|reference"/>
+        <attr name="about_libraries_title_description" format="color|reference"/>
+        <attr name="about_libraries_text_description" format="color|reference"/>
+        <attr name="about_libraries_dividerDark_description" format="color|reference"/>
         <attr name="about_libraries_title_openSource" format="color|reference"/>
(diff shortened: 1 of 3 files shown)


library/src/main/res/values/attrs.xml
```diff
@@ -4,6 +4,9 @@
     <declare-styleable name="AboutLibraries">
         <attr name="about_libraries_window_background" format="color|reference"/>
         <attr name="about_libraries_card" format="color|reference"/>
+        <attr name="about_libraries_title_description" format="color|reference"/>
+        <attr name="about_libraries_text_description" format="color|reference"/>
+        <attr name="about_libraries_dividerDark_description" format="color|reference"/>
```
mikepenz: I think this should be just named `about_libraries_divider_description`. What do you think?

fettlaus: In this case we need `about_libraries_divider_openSource` as well.

mikepenz: right. but I think it makes more sense and the `dark` would just be obsolete

mikepenz:
thank you

PR #297, opened 2017-02-04
Android Image Cropper added

Rainer-Lang:


library/src/main/res/values/library_android_image_cropper_strings.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<resources>
+	<string name="define_avatar_view"></string>
```
mikepenz: "_int" fehlt :D

Rainer-Lang:
changed

library/src/main/res/values/library_android_image_cropper_strings.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<resources>
+	<string name="define_int_avatar_view"></string>
```
mikepenz: oh one more thing. the identifier "avatar_view" does not make much sense for this lib. In addition it duplicates with the existing "avatar_view" library

mikepenz:
Now it's happy. AWESOME JOB 👍

Rainer-Lang:
thanks

PR #299, opened 2017-02-06
Add httpclient android lib

odin-:


mikepenz:
Thank you very much :)

I have found an issue:
https://github.com/mikepenz/AboutLibraries/pull/299/files#diff-bb9726e2506b837700f7962e581dc44fR17
This won't work. I have looked, but found that the lib is not really open source. So we can't link to a direct class of the library. (Java does only allow you to check if a class exists, not if a package is there)

odin-:
Where did you find it is not open source? Doesn't it says Apache 2.0 license here:

https://mvnrepository.com/artifact/cz.msebera.android/httpclient/4.4.1.2

and here:

https://github.com/apache/httpclient

?

odin-:
Kind of lame of me to do 5 commits btw :-). GitHub newbie...

mikepenz:
@odin- oh :D I was following the repo link inside the description, and this repo did not contain the source for the httpclient. 
Sorry for the confusion. 

Could you still please fix the classPath? It should be the path to a specific class existing in the lib, so auto-detection works (sadly there is no fast way to check if a package exists, so it has to check if the class exists)

Nah doesn't matter. It's a PR that's fine ;). the result counts! 👍

odin-:
I compiled your project in AS and exported the aar to my project to test locally, but to be honest I didn't get HttpClient to show up. 

Any ideas?

mikepenz:
@odin- as the library is an internal version, it will only automatically show up if the classpath matches. So if the auto detection can find the class. 

This is a little bit "expensive" so I will keep the found libs as shared pref per application version. you may want to count it up to re-run the auto detection

odin-:
Excellent. Worked like a charm. So that's why it's a bit slow to display the first time.

Then it's verified on my side at least.

Learned alot, thanks.

mikepenz:
@odin- theoretically you can also disable auto detection, and just provide the identifiers of the libs you want to show. Then no "auto-detection" is required. which is many times faster

odin-:
OK should I use 

libs.getLibrary("ActionBarSherlock") and so on for this?

mikepenz:
No. 

You would simply use the `LibsBuilder` and do the following:

Disable auto detection here:
https://github.com/mikepenz/AboutLibraries/blob/develop/library/src/main/java/com/mikepenz/aboutlibraries/LibsBuilder.java#L130

You pass the identifiers here:
https://github.com/mikepenz/AboutLibraries/blob/develop/library/src/main/java/com/mikepenz/aboutlibraries/LibsBuilder.java#L108

odin-:
Excellent. Thanks.

PR #307, opened 2017-03-03
gradle version and support lib version update

illuzor:
Because it conflicts with constraint layout

screenshot_1


mikepenz:
@illuzor you can exclude specific dependencies of used libraries (you then just have to define them manually with the version you want) 
http://stackoverflow.com/a/21765656/325479

mikepenz:
but beside this. thank you for the PR.

PR #312, opened 2017-03-21
Fix headers in markdown files

joshfriend:


mikepenz:
@joshfriend thank you very much.

PR #313, opened 2017-03-27
Update Android support libraries to 25.3.0

odin-:


mikepenz:
@odin- thank you. (until a new release is out, you can simply exclude the support libs from the aboutLibraries)

PR #316, opened 2017-04-03
AHBottomNavigation, MaterialDesignDimens and PhotoView libraries added.

matmedev:


mikepenz:
@matmedev they look very good thanks. if libraries are added into the AboutLibraries. the `_int` after `define` is required here: https://github.com/mikepenz/AboutLibraries/pull/316/files#diff-280db18cd776ac1a27308243479c117dR4 otherwise those libs would show up for everyone automatically. (sample here: https://github.com/mikepenz/AboutLibraries/blob/develop/library/src/main/res/values/library_actionbarsherlock_strings.xml#L4)

And the classpath should define a specific class (java sadly does not allow to check for a package if it exists)
https://github.com/mikepenz/AboutLibraries/blob/develop/library/src/main/res/values/library_actionbarsherlock_strings.xml#L17


thank you

matmedev:
OK. I will take a look at it and get back as soon as I am ready.

matmedev:
I made some changed (see commit message). I couldn't add the classpath for MaterialDesignDimens lib since there are no classes in the lib, only constants.

PR #318, opened 2017-04-21
Added HTTP Shortcuts as a reference app

Waboodoo:
I'm using your library in my project **HTTP Shortcuts**. Thanks for the great work.


mikepenz:
Very nice Thank you

PR #322, opened 2017-05-15
Updated library info and added new library files

JSierawski:
Updated files:
library_appcompat_v7_strings.xml
library_design_strings.xml
library_googleplayservices_strings.xml
library_okio_strings.xml
library_otto_strings.xml
library_photoview_strings.xml
library_recyclerview_v7_strings.xml
library_support_annotations_strings.xml

Add new library files:
library_hockeyapp_strings.xml
library_support_cardview_strings.xml
library_support_gridlayout_strings.xml
library_support_multidex.xml

### DIFF
--- a/library/src/main/res/values/library_photoview_strings.xml
+++ b/library/src/main/res/values/library_photoview_strings.xml
@@ -14,7 +14,7 @@
 	<string name="library_photoview_isOpenSource">true</string>
 	<string name="library_photoview_repositoryLink">https://github.com/chrisbanes/PhotoView</string>
 	<!-- ClassPath for autoDetect section -->
-	<string name="library_photoview_classPath">uk.co.senab.photoview.BuildConfig</string>
+	<string name="library_photoview_classPath">com.github.chrisbanes.PhotoView</string>
 	<!-- License section -->
 	<string name="library_photoview_licenseId">apache_2_0</string>
 	<!-- Custom variables section -->
--- a/library/src/main/res/values/library_photoview_strings.xml
+++ b/library/src/main/res/values/library_photoview_strings.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 
 <resources>
-	<string name="define_int_photoview"></string>
+	<string name="define_int_photoview">year;owner</string>
 	<!-- Author section -->
 	<string name="library_photoview_author">Chris Banes</string>
 	<string name="library_photoview_authorWebsite">http://chris.banes.me/</string>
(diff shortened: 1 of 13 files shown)


library/src/main/res/values/library_photoview_strings.xml
```diff
@@ -14,8 +14,10 @@
 	<string name="library_photoview_isOpenSource">true</string>
 	<string name="library_photoview_repositoryLink">https://github.com/chrisbanes/PhotoView</string>
 	<!-- ClassPath for autoDetect section -->
-	<string name="library_photoview_classPath">com.github.chrisbanes.PhotoView</string>
+	<string name="library_photoview_classPath">uk.co.senab.photoview.BuildConfig</string>
```
mikepenz: This seems to be wrong

library/src/main/res/values/library_support_annotations_strings.xml
```diff
@@ -8,11 +8,13 @@
...
     <string name="library_support_annotations_libraryName">Support Annotations</string>
     <string name="library_support_annotations_libraryDescription">The Annotation package provides APIs to support adding annotation metadata to your apps.</string>
-    <string name="library_support_annotations_libraryWebsite"></string>
+    <string name="library_support_annotations_libraryWebsite">https://developer.android.com/reference/android/support/annotation/package-summary.html</string>
     <string name="library_support_annotations_libraryVersion">23.2.1</string>
     <!-- OpenSource section -->
     <string name="library_support_annotations_isOpenSource">true</string>
     <string name="library_support_annotations_repositoryLink"></string>
+    <!-- ClassPath for autoDetect section -->
+    <string name="library_support_annotations_classPath">android.support.annotation.MainThread</string>
```
mikepenz: Well the support annotations are IDE only, So I assume this won't end up in the app?

mikepenz:
@JSierawski thank you. I have added 2 comments to the PR

library/src/main/res/values/library_photoview_strings.xml
JSierawski: Ah, you're right - sorry...this is a different PhotoView.  I should have checked more carefully.  I'll add a new file for this.

library/src/main/res/values/library_support_annotations_strings.xml
JSierawski: It does show up in the project window as 'external library' and gets auto-detected with this classpath.  Is there a reason not to include it here?

mikepenz: Well then it's fine :). I guess it might be personal preference to include or not include it :). But having it in is better than not. Thank you

JSierawski: Yes, better safe than sorry.  Thanks for the support!

JSierawski:
Hi Mike.  I added a commit to correct the PhotoView issue you mentioned.  Thanks for your help.

mikepenz:
@JSierawski sorry I was traveling. I'll check it

PR #326, opened 2017-06-24
Added few libraries

TheGlow666:
Added few libraries which i used in my app and were missing.


mikepenz:
@TheGlow666 closed again?

PR #329, opened 2017-06-30
Add DbFlow and Update MD

AllanWang:
(Currently untested. I will be trying it tomorrow)

DbFlow doesn't have any strings so it's unlikely they will include this.
And I incremented the version for material dialogs


mikepenz:
Looks fine so far. Please report after you tested :).

AllanWang:
I tested this with external string and it worked.

mikepenz:
Thank you very much

PR #330, opened 2017-07-06
Add a bunch of new libs

rubengees:
This adds some new internal definitions for libraries, I use in one of my projects.


mikepenz:
@rubengees thank you very much :)

mikepenz:
Looks nice as far as I can tell

PR #335, opened 2017-07-17
Update some versions

AllanWang:


AllanWang:
Does the About Libraries version need to be updated for the string updates? There's been quite a few changes since 5.9.6's release on May 3rd

mikepenz:
hmm I see some definitions where updated :D

PR #337, opened 2017-07-19
Add ThirtyInch to values

StefMa:
I want to add [ThirtyInch](https://github.com/grandcentrix/ThirtyInch) to AboutLibraries.

To be honest.  I have really no clue how your lib work and if everything is correct. I don't really understand the wiki...
Whatever.
The only thing I want to reach is:
If ThirtyInch is "autodetected" (via classPath I guess) it should be available in the "libraries section"...
I hope everything is set up correctly for that 👍

### DIFF
--- a/library/src/main/res/values/library_thirtyinch_strings.xml
+++ b/library/src/main/res/values/library_thirtyinch_strings.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+	<string name="define_int_ThirtyInch">year;owner</string>
+	<string name="library_ThirtyInch_author">grandcentrix</string>
+	<string name="library_ThirtyInch_authorWebsite">https://www.grandcentrix.net/</string>
+	<string name="library_ThirtyInch_classPath">net.grandcentrix.thirtyinch.TiPresenter</string>
+	<string name="library_ThirtyInch_libraryName">ThirtyInch</string>
+	<string name="library_ThirtyInch_libraryDescription">A MVP library for Android favoring a stateful Presenter.</string>
+	<string name="library_ThirtyInch_libraryVersion">0.8.0</string>
+	<string name="library_ThirtyInch_libraryWebsite">https://github.com/grandcentrix/ThirtyInch</string>
+	<string name="library_ThirtyInch_licenseId">apache_2_0</string>
+	<string name="library_ThirtyInch_isOpenSource">true</string>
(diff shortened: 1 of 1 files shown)


mikepenz:
@StefMa well the main idea was that the developer includes the string in his library, and the aboutlibraries will find that automatically. As this did not happen to frequently most definition files got included in this lib.
These contain the classPath which allows me to check if a class exists or not, if it is there the lib is included ---> proguard will preven this. 
In the worst case you can still define the libs you want to see during creation of the LibsBuilder

StefMa:
OK. Thank you for the information.
Does `aboutLibraries` works, if I add the strings to our `ThirtyInch` project? Then I would include it in our library instead in this one. I think it will be easier to maintain...

Another question
Is the `library_thirtyinch_strings.xml` correct? If I would copy it in our lib, would AboutLibrary recognize our lib?

What is `define_int_ThirtyInch`?
We don't include ThirtyInch or `AboutLibraries` in Ti. So we don't have to define `define_ThirtyInch` but `define_int_ThirtyInch`, right? 🤔

mikepenz:
@StefMa exactly the only difference you have to make ist to remove the `int` 
I do the same for example herE: https://github.com/mikepenz/FastAdapter/blob/develop/library-core/src/main/res/values/library_fastadapter_strings.xml

Yes it looks correct as of how it looks. 

well the `define_int` or `define_` string is used by the `AboutLibraries` library to check if the resource is coming from the  AboutLibraries itself, so it is not yet allowed to show it, only if the classPath results in a result. 
if the `define_` is found without the `int` it know that the resource is coming from a library and directly shows it without further checks

StefMa:
I have tested it, but it don't work.
See this PR https://github.com/StefMa/ThirtyInch/pull/1/files

The list is just empty :(
What I'm doing wrong?
bildschirmfoto 2017-07-19 um 10 56 25

mikepenz:
I will have to look at it, but can't do this right now. 

Do you have anything in the build which might remove the resource file?

StefMa:
Ahh thank you!
Yes, in the test sample we use proGuard. It have removed the unused resources. Of course...

I've fixed it now 👍 It works if I disabling the removing with `-keep public class **.R$string`

mikepenz:
@StefMa can you please try if: 

```proguard
-keepclasseswithmembers class **.R$* {
    public static final int define_*;
}
```

would be enough too?

StefMa:
Yep, that is enough.
Thank you for your help 👍

mikepenz:
You are welcome :) 

Thanks for integrating it ;)

PR #340, opened 2017-08-04
Update dependencies and remove app_name from library

AllanWang:
The main point of this commit was to remove the `app_name` string from the library module, as it isn't used and it leaks to other dependencies. 

Since I imported this into Android Studio, I decided to update other components as well.
* Update gradle wrapper to latest rc and tool to latest alpha
* Update all of your libraries
* Update build tools and sdk to 26; the build tool is out of beta now
* Go for implementation rather than compile. This avoids leaking interfaces to app projects and reduces build time. You may want to test this out to ensure that there are no issues with maven.

If you don't want to update dependencies, you may cherry pick the app_name deletion.

Another thing I wasn't sure if you wanted was that you can add the following to your build gradle:

```gradle
android {
    defaultConfig {
        resValue "string", "library_AboutLibrary_libraryVersion", project.VERSION_NAME
    }
}
```

to avoid having to manually change the version string when you update your library.
You can keep the `VERSION_NAME` under a gradle or property file so you only need to change it once.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,14 +7,14 @@ android:
   components:
     # Uncomment the lines below if you want to
     # use the latest revision of Android SDK Tools
-    # - platform-tools
     - tools
+    - platform-tools
 
     # The BuildTools version used by your project
-    - build-tools-25.0.2
+    - build-tools-26.0.1
 
     # The SDK version used to compile your project
-    - android-25
+    - android-26
 
     # Additional components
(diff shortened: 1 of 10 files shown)


AllanWang:
@mikepenz friendly ping. I've also updated the reactive libs and added RxKotlin

mikepenz:
@AllanWang looks good as far as I can tell

PR #356, opened 2017-12-18
update library_androidannotations_strings.xml

10erlosh:
add classPath

### DIFF
--- a/library/src/main/res/values/library_androidannotations_strings.xml
+++ b/library/src/main/res/values/library_androidannotations_strings.xml
@@ -12,4 +12,5 @@
     <string name="library_androidannotations_licenseId">apache_2_0</string>
     <string name="library_androidannotations_isOpenSource">true</string>
     <string name="library_androidannotations_repositoryLink">https://github.com/excilys/androidannotations.git</string>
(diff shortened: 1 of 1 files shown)


mikepenz:
@10erlosh almost :D https://github.com/mikepenz/AboutLibraries/blob/develop/library/src/main/res/values/library_actionbarsherlock_strings.xml#L17

It needs to be a specific class. Java sadly does not allow to check if a package exists. 

Remember that this is only working if you have no proguard. if you use obfuscation you need to use the identifiers as shown in linked sample in your issue

10erlosh:
so when ProGuarding and doing like that:
```
Libs libs = new Libs(activity, Libs.toStringArray(R.string.class.getFields()));
ArrayList<Library> list = libs.prepareLibraries(activity, null, null, true, true);
list.add(libs.getLibrary("androidannotations"));
list.add(libs.getLibrary("commonslang3"));
```
only "AndroidAnnotations" and "Commons Lang 3" would show up?

mikepenz:
when proguard obfuscates then onle the ones above 
+ the libs including the definition file (so like the aboutlibs itself, ...) 
+ the ones which might prevent obfuscation 
+ the ones you defined

you can exclude ones you don't like

PR #360, opened 2018-01-05
Use Gradle's `${rootProject.ext.supportLibVersion}` as Support lib version

mmrmartin:
Today AboutLibraries has hardcoded Support libraries versions in its XMLs. From the last version of AboutLibraries the library's Gradle file use version of Support library from `${rootProject.ext.supportLibVersion}` so I extracted this variable into String resource and then I put it into these XMLs.

This will let the AboutLibraries to show the currently used version of Support and Design library instead of some old one.

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -10,6 +10,8 @@ android {
         targetSdkVersion 27
         versionCode 6010
         versionName '6.0.1'
+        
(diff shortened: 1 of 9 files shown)


mikepenz:
Really great idea. :) thanks

PR #363, opened 2018-01-15
make links in special button click-able

QuadTriangle:


mikepenz:
@QuadTriangle thanks for the PR. I am not so sure if this is a good idea, there is already the `LibsListener` which you can attach via the `LibsConfiguration`

https://github.com/mikepenz/AboutLibraries/blob/develop/library/src/main/java/com/mikepenz/aboutlibraries/LibsConfiguration.java#L195

This has a click listener allowing you to launch a website.

QuadTriangle:
PR title is somewhat misleading. I actually fixed html links in alertdialog text.
```xml
<string name="aboutLibraries_description_special1_name" >links</string>
<string name="aboutLibraries_description_special1_text" >
    <![CDATA[
         website:
         <a href="https://google.com">Google</a>
         <a href="https://github.com">Github</a>
    ]]>
</string>
```

html links in `aboutLibraries_description_text` already works. and I give my app to someone to review UI, he ask me to make  links click-able. that's why I opened a PR.

feel free to close it.

mikepenz:
Oh ok I see. Could you please just for stability reasons null check the `message` findViewById (in case some OEM at some point decides to break dialogs, and not include the `message` id)

mikepenz:
Thanks

QuadTriangle:
wasn't `catch (Exception ex) {}`  enough  for this?

mikepenz:
well it would result in the same, but I prefer having it checked so it is known that this is a possibility vs just ignoring a exception

PR #376, opened 2018-02-19
Small optimizations

EmmanuelMess:
Small optimizations:
* Reduced 3 for loops in `Libs.init()`
* Reduced 2 for loop in `Libs.prepareLibraries()`if excluding (and 1 more if autodetect is enabled)
* Used `StringBuilder` in `Libs.getAutoDetectedLibraries()`


EmmanuelMess:
ping?

mikepenz:
@EmmanuelMess I am so sorry. Currently super flooded with work. will come back to it as soon as I can.

PR #377, opened 2018-02-24
Added material-about-library

daniel-stoneuk:
I've found that quite a few people use [material-about-library](https://github.com/daniel-stoneuk/material-about-library) for their about screen and launch AboutLibraries for the list of libraries used. For example, in FastHub.

* Added an internal definition for material-about-library, cheeky I know ;)

Thanks!

### DIFF
--- a/library/src/main/res/values/library_material_about_library_strings.xml
+++ b/library/src/main/res/values/library_material_about_library_strings.xml
@@ -1,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<resources>
+	<string name="define_int_material_about_library"></string>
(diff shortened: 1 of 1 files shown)


mikepenz:
@daniel-stoneuk the class path is one of the main classes of the library. basically packagename + class name (sadly java does not allow to check for the existence of a package)

mikepenz:
This library is already included by the way :) 

https://github.com/mikepenz/AboutLibraries/blob/develop/library/src/main/res/values/library_materialaboutlibrary_strings.xml

mueller-ma:
I'm using version 6.0.8, but the material-about-library doesn't show up in the used libs list.

PR #378, opened 2018-02-25
Added missing CardView dependency

Foso:
First thank you for this library :) 
But i couldn't use it without also adding CardView as an dependency


mikepenz:
Thank you very much. You are absolutely right ;)

PR #381, opened 2018-03-27
Use api instead of compile

Jawnnypoo:
Since compile is going to be removed at some point, `api` will take its place. It essentially will have the same effect here, so nothing to worry about.


mikepenz:
@Jawnnypoo there's a reason why I have not yet updated to `api` or `implementation` the script I use to publish to maven central does not support it yet (at least it did not a few weeks ago) Changing it right now will break my release and publish process :) 

I'll adjust as soon as the publishing works ;) (and I got time to verify that it works)

mikepenz:
Thank you very much!

Issue #386, opened 2018-04-14
Remove definitions for FABsMenu


flocke:
You can remove the included definitions for FABsMenu (https://github.com/jahirfiquitiva/FABsMenu) since it already includes the definitions in the library itself. As it is now I have two entries for it:

![](https://i.imgur.com/Pk0QUiM.png)

Btw. feel free to add my project andOTP (https://github.com/andOTP/andOTP) to the list of apps that use AboutLibraries.

mikepenz:
@flocke please open a PR :)

flocke:
@mikepenz sure, will do.

mikepenz:
thank you

PR #387, opened 2018-04-16
Remove definitions for FABsMenu

flocke:
As requested in #386 this removes the values file for FABsMenu since it is already included upstream.

As a bonus this also adds my project andOTP to the list of apps that use AboutLibraries.


mikepenz:
Thank you

PR #392, opened 2018-05-16
remove ProgressBar dependency in LoaderItem

mars3142:
I wanted to use a custom layout for the **LoaderItem**, but it was not a **ProgressBar**. Because of that, I removed the unused **findViewById** call. Now **every view** can be used als loading indicator.


mikepenz:
Looks good :D

PR #393, opened 2018-05-17
Change Context source in onCreateView method

klamborowski:
Parameter container is nullable and geting Context source from it can provide NullPointerException.

Example: Screen rotation. In my app I have master-detail landscape layout (list on the left and fragment on the right) and only list in portrait layout. When we are showing LibsSupportFragment in right fragment container and then we rotate device to portrait we get NPE, because container is null


mikepenz:
Can you please add the support annotations to fix the build issue?

klamborowski:
Sorry, my bad. Fixed

PR #400, opened 2018-07-05
Add JScience library definition

TiiXel:
I hope this is ok.

I didn't ask the author to include the definition directly in his library because it's not an android specific one.


mikepenz:
Thank you so much!

PR #411, opened 2018-08-27
Eclipse PAHO license added

Hulcsa:
Hi!

I added a new license, which is for Eclipse's Paho Android Service lib. It is widely used for MQTT messaging, and would be nice if your lib could show a license for it. Could you accept this pull request please?


mikepenz:
@Hulcsa thank you very much

PR #413, opened 2018-09-20
Correct PhotoView autoDetect ClassPath

ccomeaux:
This should be targeting the file found at https://github.com/chrisbanes/PhotoView/blob/master/photoview/src/main/java/com/github/chrisbanes/photoview/PhotoView.java


mikepenz:
@ccomeaux thank you soo much!

PR #415, opened 2018-10-04
Update Libs.java

harjazB:
Adding of internal libraries should be possible even when no libraries are excluded.

### DIFF
--- a/library/src/main/java/com/mikepenz/aboutlibraries/Libs.java
+++ b/library/src/main/java/com/mikepenz/aboutlibraries/Libs.java
@@ -167,12 +167,10 @@ public class Libs {
         List<Library> extern = getExternLibraries();
         resultLibraries.addAll(extern);
 
-        if(isExcluding) {
-            for (Library lib : extern) {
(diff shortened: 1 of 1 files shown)


mikepenz:
How did the issue surface? 

I'd love to reproduce the problem first which you are fixing.

harjazB:
.WithLibraries(String... libraries) doesn't work unless you call LibsBuilder#withExcludedLibraries("any_string") in LibsBuilder()

harjazB:
I meant line with condition if(isExcluding && lib != null) in 6.1.0 version when library was added to resultLibraries only when isExcluding==true

        //Now add all libs which do not contains the info file, but are in the AboutLibraries lib
        if (internalLibraries != null) {
            for (String internalLibrary : internalLibraries) {
                Library lib = getLibrary(internalLibrary);

                if(isExcluding && lib != null) {
                    resultLibraries.add(lib);
                    libraries.put(lib.getDefinedName(), lib);
                }
            }
        }

PR #417, opened 2018-10-20
You mean 6.2?

tobiasschuerg:
AboutLibraries should be referenced, not Fastadapter.


mikepenz:
@tobiasschuerg ah yes. thank you so much.

PR #435, opened 2019-07-02
Fix centering of list item in parent view

fchauveau:
Change property of scroll bar visibility in order to don't affect horizontal centering of list items.

Before (without and with scroll bar visible)

device-2019-07-02-121137
device-2019-07-02-121144

After (without and with scroll bar visible)

device-2019-07-02-121050
device-2019-07-02-121108


mikepenz:
Thank you very much. 

The build seems to fail as travis misses the sdkmanager somehow

fchauveau:
Anything I can do about this travis build failed?

mikepenz:
No it is not caused by your PR. your changes look good and I will bring them in this weekend I hope. 

Will investigate why travis is not happy with me :D

PR #440, opened 2019-08-29
Add Android FillableLoaders library definition

Moritz-David:
I hope this is ok.

I didn't ask the author to include the definition directly in his library because it's not updated since 2015.


mikepenz:
Thank you so much for your contribution

PR #456, opened 2019-12-17
Don't ignore aboutLibraries_description_text and name

dzamlo:
Before this change, the value of `aboutLibraries_description_text` and `aboutLibraries_description_name` were ignored. This fix  that.


mikepenz:
Thank you very much

PR #479, opened 2020-03-21
Improve handling of missing version in the artifact's POM

jhansen2015:
Fixes two issues related to missing version indicators in the pom itself:
1. If `artifactPom.version` is missing, use `artifactPom.parent.version` (the version sub-element of the parent element of the artifactPom) before attempting to retrieve the artifact's parent pom.
2. When creating the `Library` instance, use the original `component.displayName` instead of attempting to reconstruct it.

Also added a logger and several debug messages.

### DIFF
--- a/plugin/src/main/groovy/com/mikepenz/aboutlibraries/plugin/AboutLibrariesProcessor.groovy
+++ b/plugin/src/main/groovy/com/mikepenz/aboutlibraries/plugin/AboutLibrariesProcessor.groovy
@@ -259,23 +259,27 @@ class AboutLibrariesProcessor {
             return
         }
 
-        libraries.add(
-                new Library(
-                        uniqueId,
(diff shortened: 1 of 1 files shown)


jhansen2015:
Hi Mike,

Here are a couple fixes.  The ticket numbers (1 & 2) are from my fork.

I'm happy to make edits based on feedback.  A specific edit would be to remove the `TODO` marker and code for the `Library` object instantiation if it looks good to you.

Josh

mikepenz:
Hey. 

Thanks a ton for this PR. 

It actually conflicts with the latest changes I was working on myself right now. 
I fixed the dependency on internal APIs adjusted library resolving code. This would also fix a problem there was with parent pom resolval. 

Let me look at your changes and cherry pick parts over :)

jhansen2015:
Sounds good! Looking forward to seeing your other changes.

PR #489, opened 2020-04-11
Fix card background color

teobaranga:
It looks like the card background color was incorrectly set to be the window background color.

After making that change, the `cardBackgroundColor` attribute failed to be resolved but I'm not sure why:
```
2020-04-11 02:48:55.338 17272-17272/com.mikepenz.aboutlibraries.sample.debug E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.mikepenz.aboutlibraries.sample.debug, PID: 17272
    java.lang.UnsupportedOperationException: Failed to resolve attribute at index 0: TypedValue{t=0x2/d=0x7f03007e a=10}
        at android.content.res.TypedArray.getColor(TypedArray.java:527)
        at com.mikepenz.aboutlibraries.ui.item.LibraryItem$ViewHolder$1.invoke(LibraryItem.kt:240)
```

Instead of using that attribute, I've changed it to be `colorSurface`, which is the default color of [Material cards](https://material.io/develop/android/components/cards/).

### DIFF
--- a/library/src/main/java/com/mikepenz/aboutlibraries/ui/item/LibraryItem.kt
+++ b/library/src/main/java/com/mikepenz/aboutlibraries/ui/item/LibraryItem.kt
@@ -237,7 +237,7 @@ class LibraryItem(private val library: Library, private val libsBuilder: LibsBui
         init {
             val ctx = itemView.context
             ctx.resolveStyledValue {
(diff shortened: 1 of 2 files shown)


mikepenz:
Thank you very much for your contribution

PR #511, opened 2020-06-16
Allow excluding dependencies at build time

mgray88:
- Mapping file defines an id to map to another id to merge with

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,7 @@ custom_license_mappings.prop // allows defining the licenseId which should be us
 custom_license_year_mappings.prop // allows defining the license Year for this library (this information CANNOT be resolved from the POM file)
 custom_name_mappings.prop // allows overwriting the name of a library if the POM specifies unexpected information
 custom_author_mappings.prop // allows overwriting the authors of a library if the POM specifies unexpected information
+custom_exclusion_list.prop // allows excluding libraries by their id at build time
 ```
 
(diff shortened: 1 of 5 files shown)


mgray88:
This is probably not the best way to do this, however I wanted to get the PR started. I am no gradle expert so I'm having trouble getting the app to use the locally built plugin.

mikepenz:
oh I am also not a gradle expert. what I usually do: 

1.) change the version in `plugin/build.gradle` to a higher version
2.) use `uploadArchives` to install it locally (see picture below)
3.) use the new version in the global `build.gradle` for the plugin

This usually works for me :) 

image

mgray88:
Thinking about this a little more, merging dependencies would essentially be excluding them, as merging a child dependency into a parent dependency means you don't really care about the child dependency info right? I think essentially what I'm going to do is add an exclusion list config file and skip the dependencies in that file

mikepenz:
@mgray88 I think for a first version that is sufficient. 

for the merging part. I believe that kind of overlaps with the multi license per library feature. as merging may also mean there are different libraries in the too-be-merged libraries. or other fields which may differ based on the modules

mgray88:
True. I'm not sure how easy it would be to handle that automatically though, it seems more like something that falls under your disclaimer `Careful optimisation and review of all licenses` :)

mikepenz:
Thanks again a lot for this PR :)

Issue #527, opened 2020-09-14
Custom library entry with a custom license


corootine:
## About this issue

I'm using a library that has another library embedded into it and therefore is part of the original source code.
However, we're required to include the license for this embedded library as well. Naturally, this one does not get auto detected.
Is there a possibility to define a custom library entry, preferably through some configuration file rather than source code.
Additionally, this embedded library uses a custom license that cannot be resolved by the AboutLibraries. 
Can I add a custom license entry in some way?

## Details
- [8.3.0] Used library version
- [29.0.3] Used gradle build tools version
- [4.0] Used tooling / Android Studio version

## Checklist

- [x] Searched for [similar issues](https://github.com/mikepenz/AboutLibraries/issues)
- [x] Checked out the [sample application](https://github.com/mikepenz/AboutLibraries/tree/develop/app)
- [x] Read the [README](https://github.com/mikepenz/AboutLibraries/blob/develop/README.md)
- [x] Checked out the [CHANGELOG](https://github.com/mikepenz/AboutLibraries/releases)
- [x] Read the [MIGRATION GUIDE](https://github.com/mikepenz/AboutLibraries/blob/develop/MIGRATION.md)

mikepenz:
Hey. 

Sure that's possible, you can do this solely by providing the meta information you want to show. consisting of:

- library
- license
- license description

You can find a dummy setup I pushed for you here:
https://github.com/mikepenz/AboutLibraries/commit/b38e6803c17a7033fc0921dd6755ae01bd44036e

corootine:
Hey, thanks for the reply! I manged to figure it out in the meantime by inspecting and debugging the code. I think it would be worthy to put this information in the README.md. What do you think? Something along the lines:

`
You can provide custom entries for libraries and licenses through string resources. In order to define a library, you have to add following identifiers:

  <string name="define_plu_mylib" translatable="false">owner</string>
  <string name="library_mylib_author" translatable="false">author</string>
  <string name="library_mylib_libraryName" translatable="false">libName</string>
   .... and other fields that are required 
`

mikepenz:
@corootine more than happy to take a PR to enhance the README for this :)

corootine:
I'll take it up sometime this week. :)

PR #536, opened 2020-10-04
Added sections on custom libraries and licenses to the README.md

corootine:
This is a follow-up on #527


mikepenz:
@corootine thank you very much

PR #552, opened 2020-11-06
Reorganize into composite build, for debugging

rpavlik:
By request :)  Took a little bit of doing, tried to do it without moving files but that ended up not working. The gist of it is that the consuming build has to "includeBuild" the plugin/library build, so I moved all the libraries and the plugin to plugin-build.  Useful guide is here: https://ncorti.com/blog/gradle-plugins-and-composite-builds

and for proof, here's a breakpoint in the plugin:

screenshot 106

You can debug by right-clicking on a task in the "Gradle" pane of the IDE, etc.

I only tried basic builds - didn't try any of your fancy release stuff.


rpavlik:
ah, looks like I goofed up the file permissions on the gradle wrapper because I'm on Windows right now.

mikepenz:
@rpavlik thank you so much for this PR it's amazing. and really works great. 

I've looked into the chances and identified that we should not move the libraries itself, and opened a PR where I reverse parts of this: https://github.com/mikepenz/AboutLibraries/pull/555 (e.g. keep library, ... in the original location)

This makes it significant easier to review

PR #587, opened 2021-01-13
Escape apostrophe for AMD license

akiller:
If the apostrophe is not escaped the project will fail to build due to improperly escaped XML being generated as below:

```xml
<string name="license_AMDPLPA_licenseName" translatable="false">AMD's plpa_map.c License</string>
```

With this fix it should generate an escaped equivalent:

```xml
<string name="license_AMDPLPA_licenseName" translatable="false">AMD\'s plpa_map.c License</string>
```


mikepenz:
thanks @akiller

PR #591, opened 2021-01-17
Add SimpleSettings Lib to featured projects

marcauberer:
Hi,
just added my project to the featured projects list.

### DIFF
--- a/README.md
+++ b/README.md
@@ -377,6 +377,7 @@ Additional dependencies can be provided via this plugins API to extend and provi
 * [MyLife - Journal](https://play.google.com/store/apps/details?id=com.pi143.mylife)
 * [LibreAV](https://github.com/projectmatris/antimalwareapp)
 * [Honda RoadSync](https://play.google.com/store/apps/details?id=com.honda.ms.dm.sab)
+* [SimpleSettings Library](https://github.com/marcauberer/simple-settings)
 
 # Developed By
 


mikepenz:
@marcauberer thank you very much. 

Looks like an interesting project. I looked through the docs but have a major question.

It looks you already use the latest AboutLibraries version: https://github.com/marcauberer/simple-settings/blob/main/simplesettings/build.gradle#L48-L50
(Which sadly was not fully published yet due to maven central issues) 

And you provide the ability to setup the libs click listener:
https://github.com/marcauberer/simple-settings/blob/4a9b9b09b64845f38e9663f2245f521ada2ee60a/simplesettings/src/main/java/com/chillibits/simplesettings/clicklistener/LibsClickListener.kt#L15

But sadly I can't find any mention or documentation about how end users may use the features of the `AboutLibraries` fully. E.g. there is no note on how to setup the plugin in the build for dependencies to be detected: https://github.com/mikepenz/AboutLibraries#gradle-plugin

marcauberer:
Hi @mikepenz. Thanks for merging!
There is a [wiki page](https://github.com/marcauberer/simple-settings/wiki/SimpleLibsPreference), which describes the setup of the gradle plugin etc.
Maybe I should connect the wiki to the readme better, so developers can get into it more easily ...

mikepenz:
Oh I missed that one when searching. Thansk

PR #592, opened 2021-01-17
Fix CC licenses

rzimmer:
Hi @mikepenz,

first thanks a lot for `AboutLibraries` and all the effort you put into it!

We use `org.reactivestreams:reactive-streams:1.0.3` in our project, which is published under CC0-1.0. Although there are library definitions for it already, they were not recognized and thus the license was opened via `spdx.org` instead of being displayed in-app.

I discovered that it might be because of typos in some of the license values files, which I fixed.

Next I added `aboutLibsId` to `CC0_1_0` to have the mapping applied automatically during build-time. This lead to `build/generated/aboutlibraries/debug/res/values/license_cc0_10_strings.xml` being generated with the extended information from the library definitions files.

```bash
$ ./gradlew exportLibraries | grep reactive
reactive-streams;org.reactivestreams:reactive-streams:1.0.3;[CC0_1_0]
```

Still I had to add `org_reactivestreams__reactive_streams:cc0_10` to `config/custom_license_mappings.prop` to make it finally work in in my working copy of `AboutLibraries`. This might be intended, though :thinking:


mikepenz:
@rzimmer thank you very much for the PR. 

You are right that we sadly still had a typo in these more uncommon license file names :/ 

I have double checked with the latest version of the library. (which made the library more flexible in terms of casing) and it will properly include the license. Please ensure to do full clean to enforce regeneration of the data.

PR #596, opened 2021-01-23
Fixes for CC0-1.0, CC-BY-2.0, CC-BY-3.0 + CC-BY-4.0

rzimmer:
This PR finally fixes auto-detection for CC0-1.0, so no custom license mapping is needed anymore (tested with `org.reactivestreams:reactive-streams:1.0.3'`).

While I was at it, I also made file names and identifiers for CC-BY-2.0, CC-BY-3.0 + CC-BY-4.0 consistent with the other available license resources. Because I don't know any libraries using these ones, I couldn't test the auto-detection, though.

Hopefully this helps in making _AboutLibraries_ even better :v:


mikepenz:
Thank you so much. that's awesome!

PR #650, opened 2021-05-22
Add android:exported attribute

OEssHRERLr770rGGLOH0:
I believe I am getting Manifest Merger errors because of the missing attribute in this lib.

Please see [Android Dev documentation](https://developer.android.com/about/versions/12/behavior-changes-12#security):

> If your app targets Android 12 and contains activities, services, or broadcast receivers that use intent filters, you must explicitly declare the [android:exported](https://developer.android.com/guide/topics/manifest/activity-element#exported) attribute for these app components.

Setting the attribute to "false" would also fix Android 12 compatibility issues. Not sure which is preferable with this lib.

### DIFF
--- a/aboutlibraries/src/main/AndroidManifest.xml
+++ b/aboutlibraries/src/main/AndroidManifest.xml
@@ -2,6 +2,7 @@
     package="com.mikepenz.aboutlibraries">
 
     <application>
-        <activity android:name="com.mikepenz.aboutlibraries.ui.LibsActivity" />
+        <activity android:name="com.mikepenz.aboutlibraries.ui.LibsActivity"
+            android:exported="true"/>
     </application>
 </manifest>


mikepenz:
Thank you for the @OEssHRERLr770rGGLOH0 according to the linked android doc it would only be required for things iwth a intent filter which this activity does not define. 

do you have some additional log on the failure?

OEssHRERLr770rGGLOH0:
First, let me make it clear that the errors happened when targeting Android S.

```
android {
    compileSdkVersion("android-S")
    buildToolsVersion "30.0.3"

    defaultConfig {
        ...
        minSdkVersion 23
        targetSdkVersion("S")
        ...
}
```

> Thank you for the @OEssHRERLr770rGGLOH0 according to the linked android doc it would only be required for things iwth a intent filter which this activity does not define.

You're right the language makes it sound like it.

> do you have some additional log on the failure?

Not much to go by unfortunately. 

But I had 4 identical Merging Errors: 

`Error: Apps targeting Android 12 and higher are required to specify an explicit value for `android:exported` when the corresponding component has an intent filter defined. See https://developer.android.com/guide/topics/manifest/activity-element#exported for details. My_App.app main manifest (this file)`

I got rid of one by adding the attribute to the Activity in my app. 

Then, when looking at a successful Merged Manifest (after downgrading SDK target version to 30) I found 3 libraries that were missing the `android:exported` attribute in activities declared in their Manifest.xml file, including this library. Considering I have only one Activity in my app, 3 errors remaining, and 3 libraries missing the `android:exported` attribute, I thought that was probably the cause of the errors.

I could confirm by trying to replicate the bug with and without this library's dependency to confirm if that helps.

mikepenz:
@OEssHRERLr770rGGLOH0 thanks for the answer. 

Without making any other changes, would the project compile and run if you add:
```
<activity android:name="com.mikepenz.aboutlibraries.ui.LibsActivity"
            android:exported="true"/>
``` 
to your AndroidManifest.xml? The merger would take care of the export in that case. 
If it will work , then I guess the documentation of Android is written wrong. if not, please have a look out for a library with an activity including intent filters

OEssHRERLr770rGGLOH0:
> <activity android:name="com.mikepenz.aboutlibraries.ui.LibsActivity"
>             android:exported="true"/>

I am still getting the 3 errors with this added to my Manifest.xml file.

mikepenz:
This would indicate that the error is with a different library. hopefully google improves the error messaging for those situations further to better indicate which lib is the culprint

OEssHRERLr770rGGLOH0:
Well, thanks for pointing me in the right direction. Adding library activities to my Manifest.xml file help me debug this. The issue wasn't caused by this library at all so I'm closing this PR.

For the curious, the issue was caused by 3 activities included in the `androidx.test:core` library version `1.3.0`. Upgrading to version `1.4.0-beta01` fixed the issue.

mikepenz:
Glad this helped :) 

Have a nice day

PR #735, opened 2022-02-26
Fix missing licenses in custom libraries

penn5:


mikepenz:
Thank you @penn5 can you please provide some additional description. 

Also reproduction steps on how I can test this?

penn5:
Honestly I didn't use this code, I just noticed that the code won't add the additional licenses if a license is required by a custom library that isn't merged and this license is not included in any other library.

mikepenz:
Thank you very much for the PR

PR #740, opened 2022-03-24
Provide color params for custom color system when using predefined co…

hvsimon:
…mpose ui


hvsimon:
Does it also need to be add in `jvm/Main` ? I am not familiar with multiplatform

mikepenz:
Thank you very much for your first PR! 🥳 
Expanded on your code to align it more with what compose does for example for `Button`. 

Beyond that, I updated the JVM section so it also benefits from this new APi.

Issue #751, opened 2022-05-09
Licence Dialoge wrong text color


Nailik:
## About this issue

When the App uses Dark Mode the text on the Dialog remains black

## Details
- [10.1.0 ] Used library version

PR #752, opened 2022-05-09
fixed HtmlText TextView color

Nailik:
use the contentColor of the colors parameter for the Android TextView to display HTML Text

closes https://github.com/mikepenz/AboutLibraries/issues/751

### DIFF
--- a/aboutlibraries-compose/src/androidMain/kotlin/com/mikepenz/aboutlibraries/ui/compose/Libraries.kt
+++ b/aboutlibraries-compose/src/androidMain/kotlin/com/mikepenz/aboutlibraries/ui/compose/Libraries.kt
@@ -122,7 +122,7 @@ fun Libraries(
                         ) {
                             HtmlText(
                                 library.licenses.firstOrNull()?.htmlReadyLicenseContent.orEmpty(),
-                                colors.contentColor.toArgb(),
+                                colors.contentColor,
                             )
                         }
                     },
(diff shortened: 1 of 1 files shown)


mikepenz:
Thank you for the PR.

aboutlibraries-compose/src/androidMain/kotlin/com/mikepenz/aboutlibraries/ui/compose/Libraries.kt
```diff
@@ -133,10 +135,10 @@ fun Libraries(
 }
 
 @Composable
-fun HtmlText(html: String, modifier: Modifier = Modifier) {
+fun HtmlText(html: String, color: Int, modifier: Modifier = Modifier) {
```
mikepenz, line 138: @Nailik not 100% certain, but wouldn't it be better to pass in the `Color` here, and do the `toArgb` conversion only within the `setTextColor`? (as I suppose the `toArgb` is only required within the `View` world)

Nailik: would probably also work, but since the html view is not reused anywhere and toArgb is only invoked when the HtmlText is shown i don't see a problem here.
Actually how you like it, i can change it real quick.

Nailik: i changed it because then the usage of HtmlText feels more composable like

mikepenz: I believe it would be cleaner. also while I know it's only used here, it may also be advised to provide a default value to the color for the web view, and have it ordered behind the modifier.

Nailik: done, i also added the parameter names to the call (both) because the one for the color is needed after reordering

mikepenz:
Thank you very much again for this PR

PR #757, opened 2022-05-14
Introduce support for custom header in compose UI

Nailik:
this feature enables to use the LazyListScope to add a custom sticky header by using `stickyHeader` or a non sticky one by adding a custom `item` inside the header block.

It also adds an ExampleHeader to the compose activity, therefore how the padding needs to be applied has changed.
The values that were set to the default values were removed because it's not necessary.

This PR is mainly for people who want to use the compose implementation with a non sticky header because the `LazyColumn` implementation would make it hard to add an element above that scrolls with the `LazyColumn` together.


mikepenz:
Thank you very much for the PR @Nailik 

I will need to look into that a bit more first. The provided API would now only offer the ability to add headers, but I suppose footers may be similarly relevant in some cases.

PR #788, opened 2022-08-07
Update plugin implementation notes

cyb3rko:
I added a note for the newer plugin DSL as the legacy plugin application should be replaced now and became the standard.

Basically what your [plugin info page](https://plugins.gradle.org/plugin/com.mikepenz.aboutlibraries.plugin) of Gradle Plugins tells us.


mikepenz:
Thank you very much

PR #811, opened 2022-11-10
fix typo of README

TigerBeanst:


mikepenz:
Thank you for the PR, really appreciate it. the original way of writing was intentional, and is also a valid variant to provide properties to gradle. As such I'll close the PR. Thanks again though.

PR #812, opened 2022-11-14
iOS and MacOS for Compose Module Support

Syer10:
I did not update the CI github action as it looks more complicated then I normally work with.

Add iOS and MacOS support

Revert updates to Compose Multiplatform and Jetpack Compose from 1.3.0 to 1.2.1, this is done for multiple reasons
1.  Kotlin native does not support `compileOnly` so you have to use `api`, but that makes it so if the consumer is not using the Compose Multiplatform Beta, it will conflict with the current dependencies.
2.  There were no api changes in the code that the compose module uses so there is no requirement to update them

Commonize `Libraries` composable and move dialog opening to `LibrariesContainer` in the android source-set

Remove useless `OK` string from android source-set, use the `android.R` one.

Make fakeData internal

Use `produceState` with background dispatcher instead of `mutableStateOf` + `LaunchedEffect`

### DIFF
--- a/aboutlibraries-compose/build.gradle.kts
+++ b/aboutlibraries-compose/build.gradle.kts
@@ -50,17 +50,46 @@ kotlin {
     android {
         publishLibraryVariants("release")
     }
+
+    iosX64()
+    iosArm64()
+    iosSimulatorArm64()
+    macosX64()
+    macosArm64()
+
+    sourceSets {
+        val commonMain by getting
+        val commonTest by getting
+        val nonAndroidMain by creating {
+            dependsOn(commonMain)
+        }
+        val nonAndroidTest by creating {
+            dependsOn(commonTest)
+        }
+
+        listOf(
+            "jvm",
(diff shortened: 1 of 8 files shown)


mikepenz:
@Syer10 do you by coincidence also have a sample app, or would you be so kind to contribute a sample app so we can test the iOS and MacOS targets?

Syer10:
Sure I'll take a look into it, I have a few ios/mac samples in other libraries I work on and I can copy one from there.

Syer10:
I attempted a sample app here https://github.com/Syer10/AboutLibraries/tree/apple_sample
But have run into issues actually loading aboutlibraries.json from the resources, there does not seem to be a good way. I tried to take the macos and ios resource loading from https://github.com/JetBrains/compose-jb/tree/master/components/resources but it doesn't seem to work.

mikepenz:
Thank you. I'll give it a look

PR #815, opened 2022-11-23
Fix default color for HtmlText and dark theme

leinardi:
This PR fixes an issue I found when using `HtmlText` together with a Dark theme:

image

Instead of hard-coding the default color of this Composable to `Color.Black`, it now uses the `LibraryDefaults.libraryColors().contentColor`, that properly support Light and Dark themes.

### DIFF
--- a/aboutlibraries-compose/src/androidMain/kotlin/com/mikepenz/aboutlibraries/ui/compose/Libraries.kt
+++ b/aboutlibraries-compose/src/androidMain/kotlin/com/mikepenz/aboutlibraries/ui/compose/Libraries.kt
@@ -155,7 +155,11 @@ fun LicenseDialog(
 }
 
 @Composable
-fun HtmlText(html: String, modifier: Modifier = Modifier, color: Color = Color.Black) {
+fun HtmlText(
+    html: String,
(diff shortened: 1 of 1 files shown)


mikepenz:
@leinardi can you reproduce this in the sample app? 

Interesting enough the color is already passed via the `LibraryDefaults` here: https://github.com/mikepenz/AboutLibraries/blob/a34b18d0996b9aff1dbe2d8489869a5535fee8bb/aboutlibraries-compose/src/androidMain/kotlin/com/mikepenz/aboutlibraries/ui/compose/Libraries.kt#LL150C48-L150C48

So it should pick the right color 🤔

leinardi:
Yep, but in mine implementation I'm calling `HtmlText` directly since I show a custom `AlertDialog` :wink:

leinardi:
If you don't like this change, could you please remove the default argument from the Composable? This way who calls it will be forced to provide a color, instead of falling back to `Color.Black`.

mikepenz:
Thanks for the explanation. Makes a lot of sense :)

mikepenz (approved):

PR #823, opened 2022-11-24
Update GitHub Actions dependencies

mikepenz:
- update versions of used GitHub actions


lynieya:
this is twisted no wonder shit got fuc up these are how you should demonstrate transparency

PR #880, opened 2023-05-04
Failsafe on missing `title` for library

TSurkis:
## Changes
- Created a new flag for the library builder.
  - Created defaults when the flag is on: 
    - name - replaced by `"uniqueId"` which represents the package name.
    - developers - replaced by an `emptyList()` if none exist.

## Update (07.05.2023)
- Recoverability is now a default behavior, resulting in the above mentioned scenarios in case of missing data.

### DIFF
--- a/aboutlibraries-core/src/androidMain/kotlin/com/mikepenz/aboutlibraries/util/AndroidParser.kt
+++ b/aboutlibraries-core/src/androidMain/kotlin/com/mikepenz/aboutlibraries/util/AndroidParser.kt
@@ -4,9 +4,12 @@ import android.util.Log
 import com.mikepenz.aboutlibraries.entity.*
 import org.json.JSONObject
 
-actual fun parseData(json: String): Result {
+actual fun parseData(json: String, recoverable: Boolean): Result {
     try {
         val metaData = JSONObject(json)
+        val firstItem: JSONObject = metaData.getJSONArray("libraries")[0] as JSONObject
+        firstItem.remove("name")
+        firstItem.remove("developers")
 
         val licenses = metaData.getJSONObject("licenses").forEachObject { key ->
             License(
@@ -21,9 +24,9 @@ actual fun parseData(json: String): Result {
         val mappedLicenses = licenses.associateBy { it.hash }
         val libraries = metaData.getJSONArray("libraries").forEachObject {
             val libLicenses = optJSONArray("licenses").forEachString { mappedLicenses[this] }.mapNotNull { it }.toHashSet()
-            val developers = optJSONArray("developers").forEachObject {
+            val developers = optJSONArray("developers")?.forEachObject {
                 Developer(optString("name"), optString("organisationUrl"))
-            }
+            } ?: emptyList()
             val organization = optJSONObject("organization")?.let {
                 Organization(it.getString("name"), it.optString("url"))
             }
@@ -33,10 +36,11 @@ actual fun parseData(json: String): Result {
             val funding = optJSONArray("funding").forEachObject {
                 Funding(getString("platform"), getString("url"))
             }.toSet()
+            val id = getString("uniqueId")
             Library(
-                getString("uniqueId"),
+                id,
                 optString("artifactVersion"),
-                getString("name"),
+                if (recoverable) optString("name", id) else getString("name"),
                 optString("description"),
                 optString("website"),
                 developers,
@@ -53,4 +57,3 @@ actual fun parseData(json: String): Result {
     }
     return Result(emptyList(), emptyList())
 }
-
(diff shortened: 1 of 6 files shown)


aboutlibraries-core/src/androidMain/kotlin/com/mikepenz/aboutlibraries/util/AndroidParser.kt
```diff
@@ -4,9 +4,12 @@ import android.util.Log
 import com.mikepenz.aboutlibraries.entity.*
 import org.json.JSONObject
 
-actual fun parseData(json: String): Result {
+actual fun parseData(json: String, recoverable: Boolean): Result {
     try {
         val metaData = JSONObject(json)
+        val firstItem: JSONObject = metaData.getJSONArray("libraries")[0] as JSONObject
+        firstItem.remove("name")
+        firstItem.remove("developers")
```
mikepenz, line 12: I believe this was accidentally left in for testing purposes, please remove

aboutlibraries-core/src/androidMain/kotlin/com/mikepenz/aboutlibraries/util/AndroidParser.kt
```diff
@@ -4,9 +4,12 @@ import android.util.Log
 import com.mikepenz.aboutlibraries.entity.*
 import org.json.JSONObject
 
-actual fun parseData(json: String): Result {
+actual fun parseData(json: String, recoverable: Boolean): Result {
```
mikepenz, line 7: Based on the adjustments, let's not have a `recoverable` flag for this behaviour, and let it happen automatically. 

Especially as this is certainly an non-anticipated behaviour - and practically is not meant to happen (given the meta data is meant to contain that information)

This will also keep the API surface more simple, and does not require a new configuration for the exposed API.

Can you please adjust this

aboutlibraries-core/src/commonMain/kotlin/com/mikepenz/aboutlibraries/Libs.kt
```diff
@@ -26,13 +27,23 @@ data class Libs constructor(
             return this
         }
 
+        /**
+         * Don't crash on missing library data. Instead fill in:
+         * [Library.name] = use [Library.uniqueId]
+         * [Library.developers] = use [emptyList]
+         */
+        fun recoverableMissingData(recoverable: Boolean): Builder {
```
mikepenz, line 35: Please drop this function, I don't think it's required to have this behaviour configured.

aboutlibraries-core/src/androidMain/kotlin/com/mikepenz/aboutlibraries/util/AndroidParser.kt
```diff
@@ -21,9 +24,9 @@ actual fun parseData(json: String): Result {
         val mappedLicenses = licenses.associateBy { it.hash }
         val libraries = metaData.getJSONArray("libraries").forEachObject {
             val libLicenses = optJSONArray("licenses").forEachString { mappedLicenses[this] }.mapNotNull { it }.toHashSet()
-            val developers = optJSONArray("developers").forEachObject {
+            val developers = optJSONArray("developers")?.forEachObject {
```
mikepenz, line 27: `forEachObject` will already handle the null case and return an `emptyList` if the `opt` returned null, no need to add the `?` elvis operator

mikepenz (changes requested):

aboutlibraries-core/src/androidMain/kotlin/com/mikepenz/aboutlibraries/util/AndroidParser.kt
TSurkis: Removed!

aboutlibraries-core/src/androidMain/kotlin/com/mikepenz/aboutlibraries/util/AndroidParser.kt
TSurkis: Both the native and Android versions return a nullable so it's best to leave it just in case.

#### Android
```Java
@Nullable public JSONArray optJSONArray(@Nullable String name) {
    Object object = opt(name);
    return object instanceof JSONArray ? (JSONArray) object : null;
}
```

#### Native
```kotlin
internal fun JsonObject.optJSONArray(key: String): JsonArray? {
    return get(key)?.jsonArray
}
```

aboutlibraries-core/src/commonMain/kotlin/com/mikepenz/aboutlibraries/Libs.kt
TSurkis: removing :)

aboutlibraries-core/src/androidMain/kotlin/com/mikepenz/aboutlibraries/util/AndroidParser.kt
TSurkis: removing :)

aboutlibraries-core/src/androidMain/kotlin/com/mikepenz/aboutlibraries/util/AndroidParser.kt
mikepenz: Yes, however `forEachObject` will take a nullable and return an `emptyList`

mikepenz (approved):

PR #888, opened 2023-06-04
Marking header slot as composable

savvasdalkitsis:
Cannot use header slot as it is not marked composable (unless i'm missing something obvious). This should fix that


mikepenz (approved):

mikepenz:
Thank you very much for the PR. 

However the `@Composable` is not required here. Please see the sample: https://github.com/mikepenz/AboutLibraries/blob/develop/app/src/main/java/com/mikepenz/aboutlibraries/sample/ComposeActivity.kt#L93-L106

(Similar you can use the `items {}`)

savvasdalkitsis:
🤦 I hadn't realised the block was inside a lazylistscope. my bad :D

PR #963, opened 2024-02-29
Library Serializer

DatL4g:
Mark `Library` and Subclasses as `Serializable` to auto-generate a serializer.

This is useful as many Compose libraries, frameworks and methods require some kind of state saving (e.g. Decompose) and `kotlin-serialization` is the recommended way to be compatible with K2.

**Additionally:** this brings the core module a step closer to go fully multiplatform and get rid of platform dependend Parser


mikepenz:
Thank you for the PR. I'll review it as soon as possible

mikepenz (approved):

PR #966, opened 2024-03-22
Set onclick listener of SimpleLibraryItem to null if no license is known

alexbakker:
Previously, tapping on a library for which no license is known would pop up a dialog with the license of whatever library was previously bound to the view holder.


mikepenz (approved):
Thank you so much for the contribution!

PR #967, opened 2024-03-23
Sort a copy of the list of libraries to prevent a crash

alexbakker:
Previously, the following crash would occur if a ``libraryComparator`` is set:

```
java.lang.UnsupportedOperationException: Operation is not supported for read-only collection
    at kotlinx.collections.immutable.implementations.immutableList.AbstractListIterator.set(Unknown Source:4)
    at java.util.List.sort(List.java:533)
    at java.util.Collections.sort(Collections.java:210)
    at com.mikepenz.aboutlibraries.viewmodel.LibsViewModel$listItems$1$1.invokeSuspend(LibsViewModel.kt:95)
    at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
```


mikepenz (approved):
Thank you very much for the PR!

PR #972, opened 2024-04-09
New API to allow some licenses only for specific libraries

mikepenz:
- add new API to allow some licenses only for specific libraries


plugin-build/plugin/src/main/kotlin/com/mikepenz/aboutlibraries/plugin/AboutLibrariesExtension.kt
```diff
@@ -127,6 +124,20 @@ abstract class AboutLibrariesExtension {
      */
     var allowedLicenses: Array<String> = emptyArray()
 
+    /**
+     * Defines the allowed licenses for specific libraries which will not result in warnings or failures depending on the [strictMode] configuration.
+     * This is useful if some dependencies have special licenses which are only used in testing and are accepted for thsi case.
```
FooIbar, line 129: > are accepted for thsi case.

Typo here?

PR #1042, opened 2025-01-03
Migrate project to use a shared convention plugin

mikepenz:
Migrate project to use a shared convention plugin


jallal1362:
Jaran majrora

PR #1049, opened 2025-01-04
Enable config cache in samples

mikepenz:
- Enable config cache in samples


mikepenz:
Seeing this problem: https://github.com/gradle/gradle/issues/31278

Issue #1085, opened 2025-02-15
[Reproducibility issue] First-time build and subsequent builds don't produce the exact same "licenses" section


papjul:
## Steps to reproduce

Building twice in a row the same flavor of an app results in a different APK.

I'm sorry I don't have a minimum reproducible build for you to try, but my project is open source, so you can try [building it by yourself](https://github.com/breezy-weather/breezy-weather).

1. Run `./gradlew assembleBasicDebug`
2. Save the APK from `app/build/outputs/apk/basic/debug/app-basic-universal-debug.apk` somewhere outside the project as first-time.apk
3. Run `./gradlew clean && ./gradlew assembleBasicDebug`
4. Save the APK from `app/build/outputs/apk/basic/debug/app-basic-universal-debug.apk` somewhere outside the project as second-time.apk
5. Extract `aboutlibraries.json` from each APK
6. Compare them

Note: `org.gradle.parallel` is `true`, but turning it to `false` reproduces the same issue, so it's not linked.

## Expected result

The `aboutlibraries.json` files are exactly the same.

## Actual result

There are differences in the `licenses` section, leading to reproducibility issue:
- The first time `aboutlibraries.json` contains a `content` attribute for the LGPL-3.0-only, which the second time `aboutlibraries.json` does not contain.
> ![Image](https://github.com/user-attachments/assets/522bbfe2-e60e-43ed-9a02-5a063a08ed3d)
 
- The first time `aboutlibraries.json` `name` for `MIT-0` is `MIT No Attribution`, while it is `MIT-0` for the second time build. This time again, the `content` is missing from second time build.
> ![Image](https://github.com/user-attachments/assets/0ac81223-f4af-43ec-8604-561219d575e8)

*(Screenshots from IzzySoft)*

## More details

It is happening at least since v11.5.0, and still reproduces with v11.6.0.

Upstream reported issue: https://github.com/breezy-weather/breezy-weather/issues/1625
(This mentions building a different flavor first, but building the same flavor produces the same result, so it's not linked to different flavors having different libraries)

mikepenz:
@papjul thank you for the report. 

The plugin fetches the license content from SPDX and (depending on configuration) also from GitHub's API, could it be that some request may not go through? Or do you see it consistently differ from first to second?

If you build the aboutLibraries file with the CLI with its respective gradle task, do you observe the same behavior? 


And this behavior was not observed in 11.4.x?

papjul:
[...]

PR #1087, opened 2025-02-16
Fix license caching not properly checking fallback license url

mikepenz:
- Fix issue causing fallback licenses not being fetched if a prior run did store the non fallback as checked in the `remoteLicenseCache`
    - FIX https://github.com/mikepenz/AboutLibraries/issues/1085


mikepenz:
If the cache includes the key because a url was checked before, we still wanna return `null` to ensure fallback fetching occurs:
- https://github.com/mikepenz/AboutLibraries/blob/05e1f0fa2621b9127d1b9e9d2e99b7b17b92ae8b/plugin-build/plugin/src/main/kotlin/com/mikepenz/aboutlibraries/plugin/util/LicenseUtil.kt#L36

PR #1146, opened 2025-04-25
Refactor plugin to support `org.gradle.configureondemand=true`

mikepenz:
- rework plugin to properly function also when configureOnDemand is enabled
- Take significant influence by the amazing `https://github.com/cashapp/licensee` project to resolve resolution during configuration time
- replace custom parsing logic with the model builder
- remove no longer needed classes 
- include artifact folder
- some cleanup
- ensure we have proper fallback if exact variant match doesn't succeed
- add credits for licensee which helped to make the new plugin version possible
- add info re disabled config cache 
- enable configure on demand
- update aboutlibraries.json files


mikepenz:
The plugin was released as https://plugins.gradle.org/plugin/com.mikepenz.aboutlibraries.plugin/12.1.0-exp01 
waiting for feedback prior to getting this merged

Issue #1148, opened 2025-04-26
Documentation Mismatch and Bug: outputPath should be a full file path, not a directory


keta1:
### Issue Description

There's a discrepancy between the documentation and the actual implementation of the `outputPath` property in the `AboutLibrariesExtension` class. The documentation describes it as "The path to the directory where the generated meta data file will be stored", but the implementation treats it as a full file path (including filename).

### Steps to Reproduce

1. Set the `outputPath` to a directory path in the project configuration:
```kotlin
aboutLibraries {
   export {
      outputPath = file("src/commonMain/composeResources/files")
   }
}
```
2. Run the `exportLibraryDefinitions` task
3. Observe the error:

```text
type 'com.mikepenz.aboutlibraries.plugin.AboutLibrariesTask' property 'outputFile' is not writable because 'AboutLibraries/app-desktop/src/commonMain/composeResources/files' is not a file.
```

### Expected Behavior

Based on the documentation which states:

https://github.com/mikepenz/AboutLibraries/blob/a168b415a13e51ca475bf30ad51d23ad038050a3/plugin-build/plugin/src/main/kotlin/com/mikepenz/aboutlibraries/plugin/AboutLibrariesExtension.kt#L413-L431

I would expect that specifying a directory would work, and the plugin would use the default filename (`aboutlibraries.json`) in the specified directory.

### Actual Behavior

The implementation is actually expecting a full file path, including the filename. Setting a directory path causes the error mentioned above.

### Code Analysis

In `AboutLibrariesTask.kt`, the `outputFile` property is annotated with `@OutputFile`, which requires it to be a file rather than a directory:

```kotlin
@get:OutputFile
protected abstract val outputFile: RegularFileProperty
```

The `configureOutputFile` method treats the `outputPath` as a file path, not a directory path:

```kotlin
// In the findFirstPresentProvider method:
providers.gradleProperty("${PROP_PREFIX}${PROP_EXPORT_OUTPUT_PATH}")
    .map { path -> projectDirectory.file(path) },
// ...
extension.export.outputPath,
```

### Suggested Fixes

Either:

1. Fix the documentation to clarify that `outputPath` must include the filename:
```kotlin
/**
 * The full path including filename for the generated meta data file.
 *
 * This path is relative to the modules project directory.
 * ...
 */
```

2. Or modify the implementation to handle directory paths by automatically appending the default filename when a directory is provided.

keta1:
[...]

PR #1150, opened 2025-04-26
Provide variant based tasks for KMP

mikepenz:
- register variant based tasks
- add gradle plugin as dependency
- add tasks per variant for KMP 
- some smaller code cleanup
- add some refined comments
- use new variant based export API
- cleanup tasks list and remove internal tasks not relevant for users to directly use
- deprecate `generateLibraryDefinitions*` API
- some cleanup
- rename outputPath to outputFile 
- rename exportVariant to variant
- add alias variants to retain compatiblity


mikepenz:
@keta1 re https://github.com/mikepenz/AboutLibraries/issues/1148 . I thought more about it, and found a good solution which wouldn't break compatibility while allowing us to also have `outputFile`

Issue #1154, opened 2025-04-27
`12.1.0-rc03` ignores `aboutLibraries.export.excludeFields`


opatry:
I just tried updating to latest rc03 and regenerated the JSON file.
I appears I get back `funding`, `website` or `scm` nodes which were ignored in the configuration.

Here is my configuration:
```kotlin
aboutLibraries {
    collect {
        configPath = file("$rootDir/license_config")
        offlineMode = true
        fetchRemoteLicense = true
        fetchRemoteFunding = false
        includePlatform = false
    }
    android {
        registerAndroidTasks = false
    }
    export {
        outputPath = file("${projectDir}/src/main/resources/licenses_desktop.json")
        excludeFields.addAll("metadata", "funding", "scm", "associated", "website", "Developer.organisationUrl", "Organization.url")
        prettyPrint = true
    }
    license {
        strictMode = StrictMode.FAIL
        allowedLicenses.addAll("Apache-2.0", "asdkl", "MIT", "EPL-1.0", "BSD-3-Clause")
    }
    library {
        duplicationMode = DuplicateMode.MERGE
        duplicationRule = DuplicateRule.SIMPLE
    }
}
```

Works perfectly fine since many versions and was working in `12.1.0-rc01` and `12.1.0-rc02`.

opatry:
@mikepenz could be linked to this changeset maybe? https://github.com/mikepenz/AboutLibraries/commit/36fde7c1a84ccc2b0f2e023d18d21424320cc8a1#r155950044

mikepenz:
@opatry thank you for the report. And yes it's exactly that. it's fun how a `SetProperty` behaves different than a normal property, with `isPresent` resolving to true right away...

PR #1155, opened 2025-04-27
Fix exclude fields not set fully

mikepenz:
- fix exclude fields (set property is present even fi not set)
  - FIX https://github.com/mikepenz/AboutLibraries/issues/1154

### DIFF
--- a/plugin-build/plugin/src/main/kotlin/com/mikepenz/aboutlibraries/plugin/BaseAboutLibrariesTask.kt
+++ b/plugin-build/plugin/src/main/kotlin/com/mikepenz/aboutlibraries/plugin/BaseAboutLibrariesTask.kt
@@ -91,12 +91,10 @@ abstract class BaseAboutLibrariesTask : DefaultTask() {
     internal abstract val variantToDependencyData: MapProperty<String, List<DependencyData>>
 
     open fun configure() {
-        if (!excludeFields.isPresent) {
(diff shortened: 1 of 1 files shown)


plugin-build/plugin/src/main/kotlin/com/mikepenz/aboutlibraries/plugin/BaseAboutLibrariesTask.kt
```diff
@@ -91,12 +91,10 @@ abstract class BaseAboutLibrariesTask : DefaultTask() {
...
-                val config = extension.exports.findByName(variant.getOrElse(""))
-                config?.excludeFields?.orNull ?: extension.export.excludeFields.get()
-            })
-        }
+        excludeFields.set(project.provider {
+            val config = extension.exports.findByName(variant.getOrElse(""))
+            config?.excludeFields?.orNull?.takeIf { it.isNotEmpty() } ?: extension.export.excludeFields.get()
+        })
 
         if (!includeMetaData.isPresent) {
```
opatry, line 99: No need to adjust this which follows the same pattern?

mikepenz: I checked with logs. And get() would fall for them if not set, e.g. is present bring false. Different from the set property 🥲.

PR #1170, opened 2025-05-12
Improves license name matching.

StephanPartzsch:
- Adds dependencies to sample project to test license name parsing
- Improves name matching for:
   - GNU Lesser General Public License (LGPL), Version 2.1
   - Mozilla Public License Version 2.0
- Adds some meta information for:
   - ASDKL
   - CTS
- Adds license name info for: 
   -  Play Core Software Development Kit Terms of Service

### DIFF
--- a/plugin-build/plugin/src/main/kotlin/com/mikepenz/aboutlibraries/plugin/mapping/SpdxLicense.kt
+++ b/plugin-build/plugin/src/main/kotlin/com/mikepenz/aboutlibraries/plugin/mapping/SpdxLicense.kt
@@ -210,14 +210,17 @@ enum class SpdxLicense(
     LGPL_2_0_or_later("GNU Library General Public License v2 or later", "LGPL-2.0-or-later"),
     LGPL_2_1_only("GNU Lesser General Public License v2.1 only", "LGPL-2.1-only",
         customMatcher = { name, _ ->
-            Regex(
-                "GNU Lesser General Public License( \\(LGPL\\))?,? (v|Version) ?2\\.1( only)?",
-                RegexOption.IGNORE_CASE
-            ).containsMatchIn(name)
+            name.equals("GNU Lesser General Public License v2.1 only", true)
+                    || name.equals("GNU Lesser General Public License (LGPL), Version 2.1", true)
         }
     ),
     LGPL_2_1_or_later("GNU Lesser General Public License v2.1 or later", "LGPL-2.1-or-later"),
-    LGPL_3_0_only("GNU Lesser General Public License v3.0 only", "LGPL-3.0-only"),
+    LGPL_3_0_only("GNU Lesser General Public License v3.0 only", "LGPL-3.0-only",
+        customMatcher = { name, _ ->
+            name.equals("GNU Lesser General Public License v3.0 only", true)
+                    || name.equals("GNU General Lesser Public License (LGPL) version 3.0", true)
+        }
+    ),
     LGPL_3_0_or_later("GNU Lesser General Public License v3.0 or later", "LGPL-3.0-or-later"),
     LGPLLR("Lesser General Public License For Linguistic Resources", "LGPLLR"),
     Libpng("libpng License", "Libpng"),
@@ -247,11 +250,10 @@ enum class SpdxLicense(
     MPL_1_0("Mozilla Public License 1.0", "MPL-1.0"),
     MPL_1_1("Mozilla Public License 1.1", "MPL-1.1"),
     MPL_2_0("Mozilla Public License 2.0", "MPL-2.0",
-        customMatcher = { name, _ ->
-            Regex(
-                "Mozilla Public License( Version)? 2\\.0",
-                RegexOption.IGNORE_CASE
-            ).containsMatchIn(name)
+        customMatcher = { name, url ->
+            name.contains("Mozilla Public License", true)
+                    && url?.contains("mozilla.org", true) == true
+                    && url.contains("MPL/2.0", true)
         }
     ),
     MPL_2_0_no_copyleft_exception(
--- a/plugin-build/plugin/src/main/kotlin/com/mikepenz/aboutlibraries/plugin/mapping/SpdxLicense.kt
+++ b/plugin-build/plugin/src/main/kotlin/com/mikepenz/aboutlibraries/plugin/mapping/SpdxLicense.kt
@@ -208,6 +208,10 @@ enum class SpdxLicense(
     Leptonica("Leptonica License", "Leptonica"),
     LGPL_2_0_only("GNU Library General Public License v2 only", "LGPL-2.0-only"),
     LGPL_2_0_or_later("GNU Library General Public License v2 or later", "LGPL-2.0-or-later"),
+    LGPL_2_1_only("GNU Lesser General Public License v2.1 only", "LGPL-2.1-only", customMatcher = { name, _ ->
+        name.contains("GNU Lesser General Public License", true)
+                && name.contains("2.1", true)
+    }),
     LGPL_2_1_or_later("GNU Lesser General Public License v2.1 or later", "LGPL-2.1-or-later"),
     LGPL_3_0_only("GNU Lesser General Public License v3.0 only", "LGPL-3.0-only"),
     LGPL_3_0_or_later("GNU Lesser General Public License v3.0 or later", "LGPL-3.0-or-later"),
@@ -238,6 +242,10 @@ enum class SpdxLicense(
     mpich2("mpich2 License", "mpich2"),
     MPL_1_0("Mozilla Public License 1.0", "MPL-1.0"),
     MPL_1_1("Mozilla Public License 1.1", "MPL-1.1"),
+    MPL_2_0("Mozilla Public License 2.0", "MPL-2.0", customMatcher = { name, _ ->
+        name.contains("Mozilla Public License", true)
+                && name.contains("2.0", true)
+    }),
     MPL_2_0_no_copyleft_exception(
         "Mozilla Public License 2.0 (no copyleft exception)",
         "MPL-2.0-no-copyleft-exception"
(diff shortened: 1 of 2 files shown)


plugin-build/plugin/src/main/kotlin/com/mikepenz/aboutlibraries/plugin/mapping/SpdxLicense.kt
```diff
@@ -208,7 +208,6 @@ enum class SpdxLicense(
     Leptonica("Leptonica License", "Leptonica"),
     LGPL_2_0_only("GNU Library General Public License v2 only", "LGPL-2.0-only"),
     LGPL_2_0_or_later("GNU Library General Public License v2 or later", "LGPL-2.0-or-later"),
-    LGPL_2_1_only("GNU Lesser General Public License v2.1 only", "LGPL-2.1-only"),
```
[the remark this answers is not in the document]
StephanPartzsch: Sure. I thought, with the special name matching, it should be moved to the `// Special handling section`, like it is for "BSD_2_Clause" and others.

I've moved it back.

plugin-build/plugin/src/main/kotlin/com/mikepenz/aboutlibraries/plugin/mapping/SpdxLicense.kt
```diff
@@ -239,7 +238,6 @@ enum class SpdxLicense(
     mpich2("mpich2 License", "mpich2"),
     MPL_1_0("Mozilla Public License 1.0", "MPL-1.0"),
     MPL_1_1("Mozilla Public License 1.1", "MPL-1.1"),
-    MPL_2_0("Mozilla Public License 2.0", "MPL-2.0"),
```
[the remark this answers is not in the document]
StephanPartzsch: same ;)

mikepenz (approved):

plugin-build/plugin/src/main/kotlin/com/mikepenz/aboutlibraries/plugin/mapping/SpdxLicense.kt
```diff
@@ -208,7 +208,10 @@ enum class SpdxLicense(
     Leptonica("Leptonica License", "Leptonica"),
     LGPL_2_0_only("GNU Library General Public License v2 only", "LGPL-2.0-only"),
     LGPL_2_0_or_later("GNU Library General Public License v2 or later", "LGPL-2.0-or-later"),
-    LGPL_2_1_only("GNU Lesser General Public License v2.1 only", "LGPL-2.1-only"),
+    LGPL_2_1_only("GNU Lesser General Public License v2.1 only", "LGPL-2.1-only", customMatcher = { name, _ ->
+        name.contains("GNU Lesser General Public License", true)
```
mikepenz, line 212: this could accidentally match with the `GNU Lesser General Public License v2.1 or later"` license

mikepenz (changes requested):
Just noticed that the LGPL matching could wrongfully match with a different license.

plugin-build/plugin/src/main/kotlin/com/mikepenz/aboutlibraries/plugin/mapping/SpdxLicense.kt
StephanPartzsch: Sorry, you are absolutely right. I fixed this. Do you agree with the solution?

mikepenz: I think it might be better to include the term `only` as required. as `!later` might not be 100% accurate all the times.

plugin-build/plugin/src/main/kotlin/com/mikepenz/aboutlibraries/plugin/mapping/SpdxLicense.kt
```diff
@@ -239,7 +243,10 @@ enum class SpdxLicense(
     mpich2("mpich2 License", "mpich2"),
     MPL_1_0("Mozilla Public License 1.0", "MPL-1.0"),
     MPL_1_1("Mozilla Public License 1.1", "MPL-1.1"),
-    MPL_2_0("Mozilla Public License 2.0", "MPL-2.0"),
+    MPL_2_0("Mozilla Public License 2.0", "MPL-2.0", customMatcher = { name, _ ->
+        name.contains("Mozilla Public License", true)
+                && name.contains("2.0", true)
```
mikepenz, line 248: practically we have the same situation here as well. this would also match the `MPL_2_0_no_copyleft_exception` license.

mikepenz: do you have an example library which you try to fix the matching for? maybe there's an alternative rule that is more secure. like matching the url?

plugin-build/plugin/src/main/kotlin/com/mikepenz/aboutlibraries/plugin/mapping/SpdxLicense.kt
StephanPartzsch: To get it working without any drawbacks, I decided to use a regex to match the name. It is a bit harder to read, but should work much better. The other solution would be to leave the rule as it was at the beginning and add a rule for the exact (slightly) other name. What do you think?

plugin-build/plugin/src/main/kotlin/com/mikepenz/aboutlibraries/plugin/mapping/SpdxLicense.kt
StephanPartzsch: Used regex here as well.

Do you remember the commented out `implementation`s in `build.gradle.kts`? These are dependencies with the libraries and the licence names I try to match. With these you can try it out

plugin-build/plugin/src/main/kotlin/com/mikepenz/aboutlibraries/plugin/mapping/SpdxLicense.kt
```diff
@@ -239,7 +246,14 @@ enum class SpdxLicense(
     mpich2("mpich2 License", "mpich2"),
     MPL_1_0("Mozilla Public License 1.0", "MPL-1.0"),
     MPL_1_1("Mozilla Public License 1.1", "MPL-1.1"),
-    MPL_2_0("Mozilla Public License 2.0", "MPL-2.0"),
+    MPL_2_0("Mozilla Public License 2.0", "MPL-2.0",
+        customMatcher = { name, _ ->
```
mikepenz, line 250: I think matching for the url to contains `mozilla.org/MPL/2.0` is probably an easier and more secure approach in that case. (also should be cheaper than constructing a regex)

plugin-build/plugin/src/main/kotlin/com/mikepenz/aboutlibraries/plugin/mapping/SpdxLicense.kt
```diff
@@ -208,7 +208,14 @@ enum class SpdxLicense(
     Leptonica("Leptonica License", "Leptonica"),
     LGPL_2_0_only("GNU Library General Public License v2 only", "LGPL-2.0-only"),
     LGPL_2_0_or_later("GNU Library General Public License v2 or later", "LGPL-2.0-or-later"),
-    LGPL_2_1_only("GNU Lesser General Public License v2.1 only", "LGPL-2.1-only"),
+    LGPL_2_1_only("GNU Lesser General Public License v2.1 only", "LGPL-2.1-only",
```
mikepenz, line 211: looking at: https://repo1.maven.org/maven2/com/github/librepdf/openpdf-parent/1.3.9/openpdf-parent-1.3.9.pom

```
<license>
<name>GNU General Lesser Public License (LGPL) version 3.0</name>
<url>http://www.gnu.org/licenses/lgpl.html</url>
<distribution>repo</distribution>
</license>
```

unfortunately the license url can't be used here. as it doesn't include the version.

StephanPartzsch: Got the point. And when you check other versions of the lib, the license url is different too (contains the version but this seems to bit flaky).
Therefore explicit name matching is probably the most safe way, isn't it?

StephanPartzsch: Added it for version 3.0 as well (see your example pom)

plugin-build/plugin/src/main/kotlin/com/mikepenz/aboutlibraries/plugin/mapping/SpdxLicense.kt
```diff
@@ -208,9 +208,19 @@ enum class SpdxLicense(
...
+        customMatcher = { name, _ ->
+            name.equals("GNU Lesser General Public License v2.1 only", true)
+                    || name.equals("GNU Lesser General Public License (LGPL), Version 2.1", true)
+        }
+    ),
     LGPL_2_1_or_later("GNU Lesser General Public License v2.1 or later", "LGPL-2.1-or-later"),
-    LGPL_3_0_only("GNU Lesser General Public License v3.0 only", "LGPL-3.0-only"),
+    LGPL_3_0_only("GNU Lesser General Public License v3.0 only", "LGPL-3.0-only",
+        customMatcher = { name, _ ->
+            name.equals("GNU Lesser General Public License v3.0 only", true)
```
mikepenz, line 220: this is not needed. exact name matching is done even with a custom matcher. so only the `name.equals("GNU General Lesser Public License (LGPL) version 3.0", true)` is needed.

plugin-build/plugin/src/main/kotlin/com/mikepenz/aboutlibraries/plugin/mapping/SpdxLicense.kt
mikepenz: thanks!

plugin-build/plugin/src/main/kotlin/com/mikepenz/aboutlibraries/plugin/mapping/SpdxLicense.kt
mikepenz: Thanks. 

if only all would use spdx identifiers :D

mikepenz (approved):

mikepenz:
Thank you again @StephanPartzsch for the PR and for the patience to apply all the feedback.

StephanPartzsch:
You're welcome :)

PR #1256, opened 2025-11-21
Fix incorrect keep.xml syntax in withContext() documentation

smelfungus:
The `keep.xml` example in the `withContext()` documentation mistakenly adds a file extension (.json) to the resource reference. As a result, R8’s optimized resource shrinking may not properly maintain the `aboutlibraries.json` file, which can lead to crashes when calling `Libs.Builder().withContext(context).build()`.


mikepenz (approved):

mikepenz:
Thank you @smelfungus

PR #1267, opened 2025-12-12
Migrate back to normal collections and mark them stable

Goooler:
Eliminates `org.jetbrains.kotlinx:kotlinx-collections-immutable` as a transitive dependency and decreases app sizes a bit.

```
OLD: old.apk (signature: none)
NEW: new.apk (signature: none)

          │            compressed             │           uncompressed            
          ├───────────┬───────────┬───────────┼───────────┬───────────┬───────────
 APK      │ old       │ new       │ diff      │ old       │ new       │ diff      
──────────┼───────────┼───────────┼───────────┼───────────┼───────────┼───────────
      dex │   1.5 MiB │   1.4 MiB │ -15.8 KiB │     3 MiB │   2.9 MiB │ -32.5 KiB 
     arsc │ 963.6 KiB │ 963.6 KiB │       0 B │ 963.5 KiB │ 963.5 KiB │       0 B 
 manifest │   2.1 KiB │   2.1 KiB │       0 B │   7.8 KiB │   7.8 KiB │       0 B 
      res │ 459.1 KiB │ 459.1 KiB │       0 B │ 816.3 KiB │ 816.3 KiB │       0 B 
   native │    18 KiB │    18 KiB │       0 B │  36.5 KiB │  36.5 KiB │       0 B 
    asset │  26.1 KiB │    26 KiB │    -104 B │ 192.2 KiB │ 191.2 KiB │  -1,003 B 
    other │  51.7 KiB │  51.7 KiB │       0 B │ 114.1 KiB │ 114.1 KiB │       0 B 
──────────┼───────────┼───────────┼───────────┼───────────┼───────────┼───────────
    total │   2.9 MiB │   2.9 MiB │ -15.9 KiB │   5.1 MiB │     5 MiB │ -33.5 KiB 

 DEX     │ old   │ new   │ diff               
─────────┼───────┼───────┼────────────────────
   files │     1 │     1 │    0               
 strings │ 14506 │ 14448 │  -58 (+159 -217)   
   types │  4332 │  4296 │  -36 (+152 -188)   
 classes │  3349 │  3313 │  -36 (+135 -171)   
 methods │ 22673 │ 22434 │ -239 (+2480 -2719) 
  fields │ 12340 │ 12239 │ -101 (+1077 -1178) 

 ARSC    │ old  │ new  │ diff 
─────────┼──────┼──────┼──────
 configs │  162 │  162 │  0   
 entries │ 5018 │ 5018 │  0   

=================
====   APK   ====
=================

      compressed      │     uncompressed      │                                                                                            
──────────┬───────────┼───────────┬───────────┤                                                                                            
 size     │ diff      │ size      │ diff      │ path                                                                                       
──────────┼───────────┼───────────┼───────────┼────────────────────────────────────────────────────────────────────────────────────────────
  1.4 MiB │ -15.8 KiB │   2.9 MiB │ -32.5 KiB │ ∆ classes.dex                                                                              
    127 B │    +127 B │       5 B │      +5 B │ + META-INF/services/g8.r                                                                   
    127 B │    +127 B │       5 B │      +5 B │ + META-INF/services/h8.a                                                                   
          │    -127 B │           │      -5 B │ - META-INF/services/l8.r                                                                   
          │    -127 B │           │      -5 B │ - META-INF/services/m8.a                                                                   
 19.6 KiB │     -54 B │ 185.1 KiB │    -953 B │ ∆ assets/composeResources/aboutlibraries.app.generated.resources/files/aboutlibraries.json 
  5.7 KiB │     -36 B │   5.5 KiB │     -36 B │ ∆ assets/dexopt/baseline.prof                                                              
    742 B │     -14 B │     606 B │     -14 B │ ∆ assets/dexopt/baseline.profm                                                             
──────────┼───────────┼───────────┼───────────┼────────────────────────────────────────────────────────────────────────────────────────────
  1.5 MiB │ -15.9 KiB │   3.1 MiB │ -33.5 KiB │ (total)                                                                                    
```


mikepenz:
Thank you for the PR @Goooler 

Can you please explain a bit more on this? I understand removing the dependency is great, however is this the new adviced approach on stable collections? 🤔

Goooler:
Note that this is a breaking change and should be released in a major version.

mikepenz:
Screenshot 2025-12-20 at 12 05 23

Ok. I probably have to revert this. `Libs` is now reported unstable, regardless of the `stabilityConfigurationFiles` marking it stable...

mikepenz:
> Screenshot 2025-12-20 at 12 05 23
> Ok. I probably have to revert this. `Libs` is now reported unstable, regardless of the `stabilityConfigurationFiles` marking it stable...

Ok. I think I have a workaround, by also forcing `Libs` to be stable directly

mikepenz:
I don't seem to find clear guidance on this. is the stability config file shipped with the project @Goooler ? Or would that mean everyone would now have to copy paste a stability config file to their project, so `Libs` are considered stable?

Goooler:
There appears to be a bug where Compose reports are not always generated. So we can't diff the reports for now... I noticed the bug in https://github.com/JetBrains/kotlinconf-app/pull/439 too.

mikepenz:
> There appears to be a bug where Compose reports are not always generated. So we can't diff the reports for now... I noticed the bug in [JetBrains/kotlinconf-app#439](https://github.com/JetBrains/kotlinconf-app/pull/439) too.

Thanks, I somehow after re-building multiple times and adding Libs specifically got it to show stable in the report.

The remaining question is now for users of the library. if the stability is not exposed to users of the lib, and the users would have to provide a stability config file, that would not be a great experience? 

And so far I have not found a reference which would indicate that stability is kept with the published libraries. so instead, I suspect that I should add `compose` to the `core` module, and mark the respective classes stable with the annotation.

Goooler:
> The remaining question is now for users of the library. if the stability is not exposed to users of the lib, and the users would have to provide a stability config file, that would not be a great experience?

I believe once we mark the models provided by this library stable, the Compose compiler will generate stable markers for compiled code. Users should benefit from the artifacts directly without any extra configs.

mikepenz:
🤞 Let's hope. 

I can't find any documentation or reference stating this :/ as most docs are only written for app developers, not for library authors.

Goooler:
You can view the diff in #1281. I finally made reports work by adding a dummy new class.

azabost:
> That's outdated...

@Goooler Could you clarify the source of this information? I can't see any indication of this being outdated in the previously linked [documentation](https://developer.android.com/develop/ui/compose/performance/stability/fix#immutable-collections). I stumbled upon this comment by accident and it piqued my interest 🤔

Goooler:
'Outdated' means that the way immutable collections are used is recommended by tradition. The official recommend using normal collections and a stability file as preferred. And you can view my investigation in #1281.
I think they forgot updating the part of documentation.

PR #1277, opened 2025-12-19
Fix inconsistency in Libs parsing between Android and Desktop targets

### LINKED ISSUE #1276: [BUG] Inconsistency in Libs parsing between Android and Desktop targets
Hi, I noticed an inconsistency in how the `Libs` class is parsed on different targets in the AboutLibraries project.

### Android
https://github.com/mikepenz/AboutLibraries/blob/4c0a8e468525bf61b0861a430c770e1be1d53ad6/aboutlibraries-core/src/androidMain/kotlin/com/mikepenz/aboutlibraries/util/AndroidParser.kt#L34

### Desktop
https://github.com/mikepenz/AboutLibraries/blob/4c0a8e468525bf61b0861a430c770e1be1d53ad6/aboutlibraries-core/src/multiplatformMain/kotlin/com/mikepenz/aboutlibraries/util/MultiplatformParser.kt#L34

the ` optString("organisationUrl")` used in the JVM target is your extension function which returns null. but the android one returns ""(empty string) This causes the code that I'm using for desktop target, crash on android because it returns "" instead of null!
```
developer.organisationUrl?.let(openUrl) // crash on android! empty URL i is not valid!
```

even though compose module works as expected. but this can breaks custom UIs especially on android

amir1376:
closes #1276

### DIFF
--- a/aboutlibraries-core/src/androidMain/kotlin/com/mikepenz/aboutlibraries/util/AndroidParser.kt
+++ b/aboutlibraries-core/src/androidMain/kotlin/com/mikepenz/aboutlibraries/util/AndroidParser.kt
@@ -18,10 +18,10 @@ actual fun parseData(json: String): Result {
         val licenses = metaData.getJSONObject("licenses").forEachObject { key ->
             License(
                 getString("name"),
-                optString("url"),
-                optString("year"),
-                optString("spdxId"),
-                optString("content"),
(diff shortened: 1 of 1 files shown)


mikepenz:
Thank you very much for your contribution

mikepenz:
thanks for the improvement. this was merged. (had to rebase things due to the conflict)

PR #1308, opened 2026-03-04
docs: add description resource strings to keep file guidance

jlee129589:
## Summary
- When using a custom `values/aboutlibraries_description.xml`, the string resources it defines (e.g. `aboutLibraries_showLicense`, `aboutLibraries_description_name`, etc.) are also removed by optimized resource shrinking (enabled by default in AGP 9.0).
- Documents that these strings should be included in the resource keep file alongside `@raw/aboutlibraries`.

## Context
Encountered this issue after upgrading to AGP 9.0. The `aboutlibraries.json` keep rule was already documented, but the description string resources were silently stripped, causing the library description UI configuration to be lost at runtime.

## Test plan
- [x] Verified the documented keep file entries match the string resource names used in `aboutlibraries_description.xml`

### DIFF
--- a/README.md
+++ b/README.md
@@ -297,6 +297,10 @@ with the following contents:
     tools:keep="@raw/aboutlibraries" />
 ```
 
+<details><summary><b>Legacy view-based UI: keeping description resource strings</b></summary>
+
+> **Note:** This only applies to the deprecated legacy view-based UI module (`aboutlibraries`). Consider migrating to the Compose-based UI instead.
+
 If you are using a custom `values/aboutlibraries_description.xml` to configure the library description UI, the string resources it defines will also be removed by resource shrinking. Add them to the keep file as well:
 
 ```xml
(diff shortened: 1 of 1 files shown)


mikepenz:
Thank you for the PR @jlee129589 . This only affects the already deprecated legacy module, which will be removed soon. We strongly urge to update to the compose based UI instead. 

Given the deprecated status of the legacy ui please have this info only collapsed in an expendable section (so it does not take a lot of space in the README

jlee129589:
Thanks for the review @mikepenz! I've wrapped the description resource strings section in a collapsed `<details>` block and added a note that it only applies to the deprecated legacy view-based UI module.

mikepenz (approved):

mikepenz:
thank you for your contribution

jlee129589:
Thanks for merging! Happy to contribute. 🙌

PR #1335, opened 2026-04-12
chore(renovate): extend shared convention preset

mikepenz:
## Summary
- Replaces inline Renovate config with the shared `local>mikepenz/convention` preset
- Aligns this repo with https://github.com/mikepenz/agent-approver/blob/main/renovate.json

## Test plan
- [ ] Renovate dry-run / next scheduled run picks up the new preset


mikepenz:
Created in the wrong repo, see markdown-renderer instead.

PR #1345, opened 2026-04-21
feat(plugin): add includeLicenses build-time filter

### LINKED ISSUE #1306: Support specifying licenses to include in the json
We (and I believe most others) are using this library to show a screen of **open source** software licenses, which means closed-source libraries (often with proprietary licenses or no licenses) don't need to be shown anyway.

It would be great if we can specify which licenses to include in the json.

Now we can only exclude libraries by artifact IDs (which is tedious) or filter licenses programmatically after reading the json (which doesn't solve the problem that the json includes unnecessary info).

mikepenz:
## Summary

- Adds `includeLicenses` property to the `license { }` DSL block, allowing users to filter the generated JSON to only include libraries with specific licenses (by SPDX ID, name, or URL)
- When empty (default), all libraries are included — no behavior change for existing users
- Matching is case-insensitive and consistent with existing `allowedLicenses` validation logic

### Usage
```kotlin
aboutLibraries {
    license {
        includeLicenses.addAll("Apache-2.0", "MIT")
    }
}
```

Closes #1306

## Test plan

- [x] New test: `includeLicenses filters libraries by license` — verifies Apache-2.0 libs kept, MIT libs excluded
- [x] New test: `empty includeLicenses includes all libraries` — default behavior preserved
- [x] New test: `includeLicenses and exclusionPatterns work together` — both filters applied independently
- [x] Full plugin test suite passes with no regressions

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/plugin-build/plugin/src/main/kotlin/com/mikepenz/aboutlibraries/plugin/AboutLibrariesExtension.kt
+++ b/plugin-build/plugin/src/main/kotlin/com/mikepenz/aboutlibraries/plugin/AboutLibrariesExtension.kt
@@ -104,6 +104,7 @@ abstract class AboutLibrariesExtension {
             it.allowedLicenses.convention(emptySet())
             it.allowedLicensesMap.convention(emptyMap())
             it.additionalLicenses.convention(emptySet())
+            it.includeLicenses.convention(emptySet())
             it.strictMode.convention(StrictMode.IGNORE)
         }
         android {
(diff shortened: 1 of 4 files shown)


pablo861pb-cell:
<<beginOptional>> Universidad de Ejemplo License <<endOptional>>
Copyright (c) <<var;name="bullet";original="*";match=".{0,20}">> ...

pablo861pb-cell (approved):

PR #1385, opened 2026-05-26
Add `overscroll` effect parameter

langsmith:
## Summary

Adds an `overscrollEffect` option to the Compose libraries list so callers can control the `LazyColumn` overscroll behavior.

## Issue

`LibrariesScaffold`/the shared list scaffold always used the default `LazyColumn` overscroll behavior internally. Consumers had no way to disable overscroll or provide a custom `OverscrollEffect`, which leads to the stretchy overscroll effect 👇🏽

https://github.com/user-attachments/assets/05da0192-9e66-482f-9d5a-2d4239e0ed60

## Fix

- Added `overscrollEffect: OverscrollEffect? = rememberOverscrollEffect()` to the shared list APIs.
- Passed the value through M2/M3 `LibrariesContainer` and `Libraries` into `LibraryListScaffold`.
- Applied `overscrollEffect = overscrollEffect` directly to the backing `LazyColumn`.

Existing behavior remains unchanged by default. Callers can pass `null` to disable overscroll or pass a custom remembered effect 

When overscroll is set to `null` as a parameter when using `M3LibrariesContainer()` in the test app👇🏽

https://github.com/user-attachments/assets/f8b0c4e9-ba39-47b8-9246-a6a0340ed235

### DIFF
--- a/sample/shared/src/commonMain/kotlin/com/mikepenz/aboutlibraries/sample/App.kt
+++ b/sample/shared/src/commonMain/kotlin/com/mikepenz/aboutlibraries/sample/App.kt
@@ -315,7 +315,6 @@ fun App(libs: Libs?) {
                     val contrastLevel = if (settings.highContrast) ContrastLevel.High else ContrastLevel.Normal
                     if (settings.useMaterial3) {
                         M3LibrariesContainer(
-                            overscrollEffect = null,
                             libraries = libs?.let { Libs(filteredLibs, it.licenses) },
                             modifier = Modifier.weight(1f).fillMaxWidth(),
                             badges = settings.badges,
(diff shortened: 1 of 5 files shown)


sample/shared/src/commonMain/kotlin/com/mikepenz/aboutlibraries/sample/App.kt
```diff
@@ -315,6 +315,7 @@ fun App(libs: Libs?) {
                     val contrastLevel = if (settings.highContrast) ContrastLevel.High else ContrastLevel.Normal
                     if (settings.useMaterial3) {
                         M3LibrariesContainer(
+                            overscrollEffect = null,
```
mikepenz: can you reverse this one please. I'd love to keep the standard overscroll effect in the sample app

mikepenz:
Thank you very much for the contribution

sample/shared/src/commonMain/kotlin/com/mikepenz/aboutlibraries/sample/App.kt
langsmith: Oh, whoops. Snuck in there. Totally agree. I'll remove.

PR #1438, opened 2026-08-09
feat(plugin): add `mergePlatformArtifacts` to report KMP artifacts under their root coordinate

### LINKED ISSUE #1430: Resolved uniqueId does not always match declared dependencies' anymore
Starting from release 14.1.0 (and still in version 15.0.4), the collected libraries as reflected by their `uniqueId` seem to correspond to the resolved artifact instead of the root/common coordinate as is usually declared in the build script. They are usually identical, but not always.

For example, declaring `implementation("com.mikepenz:aboutlibraries-compose-core:<version>")` would result, in the generated `aboutlibraries.json`:
- with 14.0.1: `"uniqueId": "com.mikepenz:aboutlibraries-compose-core"`
- with 14.1.0: `"uniqueId": "com.mikepenz:aboutlibraries-compose-core-android"`

Was the change intentional?

The change makes it more difficult to programmatically match collected libraries with the ID of the dependencies declared in the build script.
I am using this to tag the root dependencies using automatically generated override files, so that I can highlight them and show additional useful information when listing them in my application.

mikepenz:
Closes #1430

> [!NOTE]
> Built on top of #1445, so the diff shows that PR's commits until it merges. Review the top commit only.

## Problem

Since 14.1.0 a dependency declared as `com.mikepenz:aboutlibraries-compose-core` can be reported as `com.mikepenz:aboutlibraries-compose-core-android`, which makes it hard to match generated entries against what the build script declares (breaking `config` overrides keyed on the declared id).

Gradle's resolution graph contains **both** the KMP root module — an `available-at` redirect shell, e.g. `androidx.collection:collection` — and the resolved platform artifact `androidx.collection:collection-jvm`.

Worth being precise about what goes wrong, because it is narrower than "both become entries": the default `DuplicateMode.MERGE` **already** collapses them onto a single entry today. What it does not do is pick *which* one survives — `minByOrNull { name.length }` over graph-iteration order does. So the bug is not a duplicate entry, it is a single entry named after one platform while standing for all of them.

That naming is harmless as long as nothing else on the entry contradicts it. `collect.includeTargets` (#1445) is exactly such a thing. Two-target (jvm + js) consumer, merging off, `duplicationMode = KEEP` to show the raw attribution:

```
androidx.collection:collection       targets: ["js", "jvm", "metadata"]   <- the redirect shell
androidx.collection:collection-js    targets: ["js"]
androidx.collection:collection-jvm   targets: ["jvm"]
```

Each is correct in isolation. But at the default `duplicationMode = MERGE` those three collapse, and the surviving entry has to carry the **union** of their targets — dropping the discarded siblings' targets would make the entry under-report, so a consumer filtering `"jvm" in it.targets` would silently lose the library from its jvm list. Union is the safe choice, and it leaves:

```
androidx.collection:collection-js    targets: ["js", "jvm", "metadata"]
```

An id that says one platform on an entry that stands for three.

## Solution

```kotlin
aboutLibraries {
    library {
        mergePlatformArtifacts = true
    }
}
```

The redirect shell is dropped and its module name carried onto the platform artifact, so the reported `uniqueId` is the declared coordinate and is stable:

```
androidx.collection:collection       targets: ["js", "jvm", "metadata"]
```

Note this is not the arbitrary survivor getting patched up after the fact. `collection-js` and `collection-jvm` both carry `rootModule = collection`, so they share a `uniqueId` *before* target attribution runs — the union is computed on the correct id, and `processDuplicates` is left with nothing to collapse. Id and targets agree because they were never separated.

- Detection reads `ResolvedComponentResult.variants[].externalVariant`. Only a component whose variants *all* redirect to a single target within the same group is treated as a shell — no suffix stripping, so `androidx.annotation:annotation-jvm` declared directly stays untouched.
- No additional artifact resolution; configuration-cache and project-isolation safe.
- Metadata (name, description, licenses) still comes from the resolved artifact's POM.
- Independent of `duplicationMode` / `duplicationRule`, applied before them.
- Defaults to `false` — existing output is unchanged.

`rootModule` is held outside `DependencyCoordinates`' constructor so it takes no part in `equals`/`hashCode`. It records *how* an artifact was reached, not which artifact it is; the same `g:a:v` reached through a redirect in one configuration and directly in another therefore stays one coordinate instead of being spread over two version slots and resolved twice. `cacheKey()` already ignored it for the same reason.

`LibraryPostProcessor` also deduplicates on an exact variant match, since several platform artifacts can now collapse onto the same root id within one configuration — and an exact hit that dedups to empty must not fall through to the prefix-match branch.

## Naming

Previously `mergeVariants`. Renamed because "variant" already has a settled and different meaning throughout this plugin — `filterVariants`, `includeTestVariants`, `export.variant`, `PROP_EXPORT_VARIANT` all denote an **Android build variant**. Read against those, `mergeVariants` sounds like it merges the debug and release reports. `mergePlatformArtifacts` says what actually merges, and stays distinct from the `targets` field added in #1445.

## Tests

`MergePlatformArtifactsFunctionalTest`, resolving real `androidx.collection` / `androidx.annotation` coordinates:

- default: root and platform artifact both reported (unchanged behaviour)
- enabled: only `androidx.collection:collection` remains, metadata preserved
- enabled, suffixed coordinate with no redirect in the graph: `annotation-jvm` kept as-is
- enabled, platform artifact reached before its shell: merge is independent of graph walk order
- non-multiplatform trees: byte-identical with and without the option
- **two-target (jvm + js) multiplatform consumer** with `collect.includeTargets` and `duplicationMode` left at its default: the merged entry is named for the declared root and carries the union of its targets

That last one is new. Every other test in the class runs on a `java-library` project, so none of them exercised a real multiplatform *consumer* — the case the option exists for.

`./gradlew :plugin:test` passes.

## Open question

Given the merging fixes output that is arguably wrong today rather than adding a new capability, should this stay opt-in, or default to on? Left opt-in here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -279,6 +279,10 @@ aboutLibraries {
         duplicationMode = com.mikepenz.aboutlibraries.plugin.DuplicateMode.MERGE
         // Configure the duplication rule, to match "duplicates" with
         duplicationRule = com.mikepenz.aboutlibraries.plugin.DuplicateRule.SIMPLE
+        // Report the per-platform artifacts of a Kotlin Multiplatform publication under the declared
+        // root coordinate (`androidx.collection:collection`, not `androidx.collection:collection-jvm`).
+        // Unrelated to Android build variants — see `filterVariants` for those.
+        mergePlatformArtifacts = true
     }
 }
 ```
--- a/plugin-build/plugin/src/main/kotlin/com/mikepenz/aboutlibraries/plugin/AboutLibrariesExtension.kt
+++ b/plugin-build/plugin/src/main/kotlin/com/mikepenz/aboutlibraries/plugin/AboutLibrariesExtension.kt
@@ -88,6 +88,7 @@ abstract class AboutLibrariesExtension {
             it.exclusionPatterns.convention(emptySet<Pattern>())
             it.duplicationMode.convention(DuplicateMode.MERGE)
             it.duplicationRule.convention(DuplicateRule.EXACT)
+            it.mergePlatformArtifacts.convention(false)
         }
         license {
             it.mapLicensesToSpdx.convention(true)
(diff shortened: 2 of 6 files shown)


mikepenz:
Thanks — both comments addressed in bbf0c32c.

**Walk-order dependence: real bug, good catch.** Confirmed with a failing test before fixing it. The failure mode is slightly worse than described: when the platform artifact is reached directly first (e.g. something declares `androidx.annotation:annotation-jvm`, which is visited before the `annotation` redirect shell that `collection` pulls in transitively), the target is in `seen`, never receives the override, *and* the shell is dropped — so the output keeps the `-jvm` id and loses the root entirely.

Rather than re-processing already-seen edges and replacing coordinates, the redirects are now collected during the walk and applied as a pass over the finished coordinate set. Same result, order-independent by construction, and it removes the `rootModule` parameter threading through the recursion.

Regression test added: `variants are merged regardless of the order the graph is walked in` declares `annotation-jvm` ahead of `collection` so the platform artifact is reached first. It fails on the previous commit.

**Test naming:** correct, `runExport()` sets `duplicationMode = KEEP`, so "by default" was misleading. Renamed to `root and resolved variant are both collected when merging is disabled`, and the comment now says why KEEP is what makes both entries visible.

PR #1445, opened 2026-08-13
Make it possible to trace the library origin in a multiplatform project

sergeys-opera:
When using `collect.all=true` it's impossible to know where the dependency comes from and whether it's applicable for a specific platform (in a multiplatform project). With this change the UI can filter dependencies based on their variants displaying only Android dependencies on Android (having `androidCompileClasspath` variant) and iOS dependencies on iOS (having `iosSimulatorArm64CompileKlibraries` variant). 

```
feat: add opt-in `variants` field reporting resolved Gradle configurations

Each library in the generated metadata can now carry a `variants` array listing the raw Gradle configuration names it was resolved from, e.g. `["androidCompileClasspath", "androidRuntimeClasspath"]`.

Enabled via `collect { includeVariants = true }`. It is disabled by default, and the plugin-side field is nullable so `excludeNulls()` in the JSON writer omits the key entirely when off — existing metadata files are unaffected. It can also be dropped per export via
`excludeFields.add("Library.variants")`.

Configuration provenance is gathered before the `uniqueId`-keyed deduplication that would otherwise discard it. The lookup is derived from the selected configurations rather than stored on `DependencyData`, since each coordinate is parsed exactly once and its `DependencyData` instance is shared by every configuration containing it. Values are sorted to keep the output byte-for-byte reproducible. Provenance is also unioned across duplicates collapsed by `DuplicateMode.MERGE` and across config overrides.

The field is exposed on the runtime `Library` entity and read by both the Android and multiplatform parsers.
```


mikepenz:
Thank you very much for your contribution @sergeys-opera

mikepenz:
To clarify your code will result in the buildConfigurations not variants (jvm, ...) 

any concerns from your side if we change this to `"configurations"` to more correctly reflect the content?

```
"variants": [
                "jvmCompileClasspath",
                "jvmMainCompileClasspath",
                "jvmMainRuntimeClasspath",
                "jvmRuntimeClasspath"
            ]
```

sergeys-opera:
> Thank you very much for your contribution @sergeys-opera

No problem 😄 

> To clarify your code will result in the buildConfigurations not variants (jvm, ...)

Yes, but that's exactly what you put in `collect.filterVariants`, don't you?
Here's an example of the configuration we have in our project:
```
    collect {
        all = true
        filterVariants.addAll(
            "androidCompileClasspath",
            "androidRuntimeClasspath",
            "iosSimulatorArm64CompileKlibraries",
        )
   }
```

> any concerns from your side if we change this to "configurations" to more correctly reflect the content?

I don't mind changing the output field name to `configurations`. But I am a bit concerned that the input parameter `collect.filterVariants` won't correspond to the output field `configurations` in that case.

Some additional thoughts: it would be much nicer to be able to tag the output dependencies by platform, like having `platforms=["common", "ios"]` or similar in the output. But that would require a custom mapping between the `collect.filterVariants` (AKA configurations) and the platforms hard-coded in the plugin which feels a bit brittle. Given that the user can provide custom configurations the mapping might not work at all (unless of course we use some generic heuristics like `configuration.contains("ios")`).

sergeys-opera:
@mikepenz please let me know if you have a better way of providing the platform info in the output file than what I came up with.

mikepenz:
> > Thank you very much for your contribution @sergeys-opera
> 
> No problem 😄
> 
> > To clarify your code will result in the buildConfigurations not variants (jvm, ...)
> 
> Yes, but that's exactly what you put in `collect.filterVariants`, don't you? Here's an example of the configuration we have in our project:
> 
> ```
>     collect {
>         all = true
>         filterVariants.addAll(
>             "androidCompileClasspath",
>             "androidRuntimeClasspath",
>             "iosSimulatorArm64CompileKlibraries",
>         )
>    }
> ```

Filter Variants was designed to be for `debug` vs `release` (years years ago while the project was android only) - it's true that this din't age well :/ 

```
filterVariants.addAll("debug", "release")
```

> 
> > any concerns from your side if we change this to "configurations" to more correctly reflect the content?
> 
> I don't mind changing the output field name to `configurations`. But I am a bit concerned that the input parameter `collect.filterVariants` won't correspond to the output field `configurations` in that case.
> 
> Some additional thoughts: it would be much nicer to be able to tag the output dependencies by platform, like having `platforms=["common", "ios"]` or similar in the output. But that would require a custom mapping between the `collect.filterVariants` (AKA configurations) and the platforms hard-coded in the plugin which feels a bit brittle. Given that the user can provide custom configurations the mapping might not work at all (unless of course we use some generic heuristics like `configuration.contains("ios")`).

I'll have another look. but I have some ideas

mikepenz:
Released in 15.1.0-b01 - it will now have the targets, like wasm, jvm, ... so closer to what you highlighted.

sergeys-opera:
Great, thanks! I'll try the new plugin version.

sergeys-opera:
It seems to be working pretty well, thanks for the quick fix!

sergeys-opera:
Actually, it looks like the iOS targets are missing in some libraries when `mergePlatformArtifacts = true`

mikepenz:
Do you have an example of that?

sergeys-opera:
Prior to setting `mergePlatformArtifacts = true` I had
```
        {
            "uniqueId": "androidx.lifecycle:lifecycle-viewmodel-savedstate",
            "artifactVersion": "2.11.0",
            "name": "Lifecycle ViewModel with SavedState",
            "description": "Android Lifecycle ViewModel",
            "website": "https://developer.android.com/jetpack/androidx/releases/lifecycle#2.11.0",
            "licenses": [
                "Apache-2.0"
            ],
            "targets": [
                "android",
                "iosSimulatorArm64"
            ]
        },
```
After setting this flag:
```
        {
            "uniqueId": "androidx.lifecycle:lifecycle-viewmodel-savedstate",
            "artifactVersion": "2.11.0",
            "name": "Lifecycle ViewModel with SavedState",
            "description": "Android Lifecycle ViewModel",
            "website": "https://developer.android.com/jetpack/androidx/releases/lifecycle#2.11.0",
            "licenses": [
                "Apache-2.0"
            ],
            "targets": [
                "android"
            ]
        },
```

mikepenz:
Having a challenging time to reproduce this in the sample: https://github.com/mikepenz/AboutLibraries/commit/6404aef286e62338089fd6af2179e864bc6e8181

sergeys-opera:
Here's the reproducer: https://github.com/mikepenz/AboutLibraries/commit/027b24a62cbef4c8876e05be8b5e13ac19212dc7
Apparently it was important to add `filterVariants` too.

mikepenz:
@sergeys-opera a new b02 is being built right now. This should also handle this case 🙏

